PR #2, opened 2026-06-20
PR2 — Tree Model & Content Renderer (T-6…T-12)

smarzban:
## What

Second layer-based PR executing `specs/file-viewer/plan.md`, tasks **T-6…T-12**, each built test-first (red-green-refactor), repo green between every task.

### Tree Model
- **T-6** — gitignore-aware **lazy** enumeration via the `ignore` crate (immediate children per directory, depth-first on expand); hides gitignored entries (AC-4) and `.git`, shows dotfiles, dirs before files; bounded by root, no node escapes it (AC-N5); read-only (AC-N1); recursive expand (AC-3).
- **T-7** — `set_show_ignored` (AC-5), `set_changed_only` (AC-6 — keep changed files + the dirs leading to them, auto-expanded; restore on toggle), `set_status` for per-node markers (AC-7, tree side).
- **T-8** — perf guard: launch interactive in <1s on a 10k-file tree (AC-22), via lazy enumeration.

### Content Renderer
- **T-9** — `classify`: **bounded** read (≤1 MB, never slurps a hostile file, AC-N1); NUL → `Binary` placeholder, no raw bytes (AC-12); ≥1 MB or ≥5000 lines → bounded preview + notice (AC-13).
- **T-10** — `to_text`: strips cursor/screen-control escapes (keeps SGR), ingests via `ansi-to-tui` → ratatui `Text` carries no terminal-control ops, so content can't move the cursor or clear the screen regardless of source (AC-27).
- **T-11** — `render`: delegates to injected markdown/diff/syntax commands (glow/delta/bat), feeding **untrusted content on stdin** (no injection; written on a thread to avoid pipe deadlock); output re-neutralized via `to_text`. Missing/failed renderer → plain-text fallback (AC-24) + capability-naming notice (AC-25). (AC-8/9/10)
- **T-12** — perf guard: in-process classify+ingest of 1 MB <300ms (AC-23 in-process bound; external render is off-thread in T-19).

## Tests
All green (~36 new assertions across `tree`, `tree_filters`, `tree_perf`, inline `render`, `render_escape`, `render_delegate`, `render_perf`). Release binary builds. Renderer/temp-dir seams keep every test hermetic (no glow/delta/bat dependency).

## Security notes (untrusted file bytes)
Content is the primary trust boundary: bounded reads cap memory, binary files never emit raw bytes, escape sequences are neutralized, and external renderers receive content only via stdin. Consistent with PR1's untrusted-repo hardening.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,7 @@
 pub mod context;
 pub mod git;
 pub mod root;
+pub mod tree;
 pub mod view_policy;
 
 /// Entry point invoked by the binary. Wires the components and runs the event loop.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@
 
 pub mod context;
 pub mod git;
+pub mod render;
 pub mod root;
 pub mod tree;
 pub mod view_policy;
--- a/src/render.rs
+++ b/src/render.rs
@@ -122,12 +122,14 @@ pub fn render(
     prepared: &Prepared,
     mode: ViewMode,
     raw_diff: Option<&str>,
+    file_name: Option<&str>,
 ) -> (Text<'static>, Option<String>) {
+    let name = file_name.unwrap_or("");
     // A diff is derived from git, not from the file's bytes, so it renders even for a
     // deleted or binary file (AC-9) — never short-circuit it to the binary placeholder.
     if mode == ViewMode::Diff {
         let diff = raw_diff.unwrap_or("");
-        return delegate(&renderers.diff, diff, mode, renderers.timeout, None);
+        return delegate(&with_name(&renderers.diff, name), diff, mode, renderers.timeout, None);
     }
 
     // Content modes: a binary file shows a placeholder, never raw bytes (AC-12).
@@ -138,15 +140,30 @@ pub fn render(
     };
 
     match mode {
-        ViewMode::RenderedMarkdown => {
-            delegate(&renderers.markdown, content, mode, renderers.timeout, base_notice)
-        }
-        ViewMode::SyntaxContent => {
-            delegate(&renderers.syntax, content, mode, renderers.timeout, base_notice)
-        }
+        ViewMode::RenderedMarkdown => delegate(
+            &with_name(&renderers.markdown, name),
+            content,
+            mode,
+            renderers.timeout,
+            base_notice,
+        ),
+        ViewMode::SyntaxContent => delegate(
+            &with_name(&renderers.syntax, name),
+            content,
+            mode,
+            renderers.timeout,
+            base_notice,
+        ),
         ViewMode::RawContent => (to_text(content), base_notice),
         ViewMode::Diff => unreachable!("handled above"),
     }
+}
+
+/// Substitute the `{name}` placeholder in a renderer command with the selected file name,
+/// so a stdin-fed renderer (e.g. `bat --file-name={name}`) can still infer the language —
+/// keeping the secure stdin design while enabling syntax highlighting (AC-10).
+fn with_name(command: &[String], name: &str) -> Vec<String> {
+    command.iter().map(|arg| arg.replace("{name}", name)).collect()
 }
 
 /// Run a renderer over `input`, ingesting its output; on missing/failed/timed-out renderer
@@ -309,11 +326,14 @@ fn strip_terminal_control(raw: &str) -> String {
                 Some(_) => i += 2, // ESC + single (e.g. ESC c reset) → drop both
                 None => i += 1,    // lone trailing ESC → drop
             }
+        } else if bytes[i] == 0xc2 && matches!(bytes.get(i + 1), Some(0x80..=0x9f)) {
+            // A C1 control codepoint (U+0080–U+009F, e.g. U+009B = CSI) encoded in UTF-8;
+            // some terminals act on these, so drop the whole 2-byte sequence.
+            i += 2;
         } else {
             // Drop other C0 control bytes (BEL/BS/CR/FF/VT/…) and DEL, which can still
             // ring the bell, backspace, or carriage-return to overwrite/spoof a line.
-            // Keep only newline and tab. C1 bytes (0x80–0x9f) are UTF-8 continuation
-            // bytes here and are left to the final lossy decode.
+            // Keep only newline and tab.
             let b = bytes[i];
             let is_c0_control = b < 0x20 && b != b'\n' && b != b'\t';
             if !is_c0_control && b != 0x7f {
--- a/src/render.rs
+++ b/src/render.rs
@@ -129,8 +129,8 @@ pub fn render(
     // A diff is derived from git, not from the file's bytes, so it renders even for a
     // deleted or binary file (AC-9) — never short-circuit it to the binary placeholder.
     if mode == ViewMode::Diff {
-        let diff = raw_diff.unwrap_or("");
-        return delegate(&with_name(&renderers.diff, name), diff, mode, renderers.timeout, None);
+        let (diff, notice) = cap_preview(raw_diff.unwrap_or(""));
+        return delegate(&with_name(&renderers.diff, name), &diff, mode, renderers.timeout, notice);
     }
 
     // Content modes: a binary file shows a placeholder, never raw bytes (AC-12).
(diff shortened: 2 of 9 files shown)


smarzban:
## Review Gate

Findings: 11 total — 0 critical · 0 high · 0 medium · 8 low · 3 info.
_Reviewed by:_ holistic + lens-security + tools · models: glm-5.2:cloud, gpt-5.5, claude-opus-4-8, deterministic

### Advisory (non-blocking)
- ⚪ **[LOW]** \[test-coverage\] MAX\_RENDER\_OUTPUT 16MB capture cap is untested — `src/render.rs:277` · 1/3 models _(test-coverage)_
  The reader thread bounds captured stdout at MAX\_RENDER\_OUTPUT to keep a spewing renderer from exhausting memory, but no test exercises a renderer that emits \>16MB; the cap could regress unnoticed.
  _Fix:_ Add a test with a renderer that prints unbounded output and assert capture is bounded and the fallback/return still completes.
- ⚪ **[LOW]** \[concurrency\] run\_renderer can block for up to 2× timeout on a wedged renderer — `src/render.rs:286` · 1/3 models _(concurrency)_
  rx.recv\_timeout(timeout) waits the full timeout for stdout EOF, then wait\_bounded(child, timeout) waits a second full timeout for exit. A renderer that closes stdout then hangs in cleanup blocks \~2×timeout. render() is synchronous in this PR (off-threading is deferred to T-19 per render\_perf.rs), so if it is wired onto the UI thread before T-19 a wedged delegate freezes the UI for double the intended bound.
  _Fix:_ Give wait\_bounded a small fixed grace (e.g. 1s) rather than reusing the full timeout — once stdout closes the process should exit near-instantly, so a second full timeout is unnecessary latency.
- ⚪ **[LOW]** \[test-coverage\] OSC and C1 escape-neutralization paths (AC-27) are untested — `src/render.rs:354` · 1/3 models _(test-coverage)_
  strip\_terminal\_control has distinct branches for OSC (ESC \] ... BEL/ST) and UTF-8 C1 controls (U+0080-U+009F, e.g. U+009B CSI), both security-relevant for AC-27, but render\_escape.rs only exercises CSI, C0, and SGR. A regression in the OSC/C1 branches would pass the suite silently.
  _Fix:_ Add cases asserting OSC (\\x1b\]0;title\\x07) and C1 CSI (U+009B) are stripped with no ESC/control byte surviving to\_text.
- ⚪ **[LOW]** \[security\] Bidi/invisible Unicode format chars not neutralized in content viewer — `src/render.rs:382` · 1/3 models _(security)_
  strip\_terminal\_control removes C0/C1/DEL and ESC sequences but keeps multi-byte Unicode format chars like U+202E RIGHT-TO-LEFT OVERRIDE (0xe2 0x80 0xae) and zero-width/joiner chars; these survive into ratatui spans and are emitted to the terminal in logical order. On a bidi-honoring terminal this lets an untrusted file/diff display text in a different order than its bytes (Trojan-Source), which directly subverts a tool whose purpose is reviewing code and diffs. AC-27 only covers cursor/screen control, leaving this deception vector open.
  _Fix:_ Strip or visibly escape Unicode bidi-control (U+202A–U+202E, U+2066–U+2069), zero-width (U+200B–U+200D, U+FEFF) and other Cf format codepoints in to\_text/strip\_terminal\_control so displayed order matches file bytes.
- ⚪ **[LOW]** \[error-handling\] read\_to\_end failure returns an empty Full pane instead of a placeholder — `src/render.rs:62` · 1/3 models _(error-handling)_
  If File::open succeeds but the bounded read fails mid-stream, classify returns Prepared::Full { text: String::new() }, showing a silently empty pane. The module's own docstring (lines 56-58) states an unreadable file should yield the Binary placeholder 'not a misleading empty pane' — this path contradicts that intent.
  _Fix:_ Return Prepared::Binary (or a dedicated notice) on read\_to\_end error, matching the open-failure branch, so a read error never renders as an empty content pane.
- ⚪ **[LOW]** \[correctness\] Large invalid UTF-8 files bypass binary classification — `src/render.rs:73` · 1/3 models _(correctness)_
  For files at or above the byte cap, any invalid UTF-8 in the first megabyte is decoded lossily and rendered, while the same bytes in a smaller file classify as binary. That makes binary handling depend on file size and can show a lossy preview instead of the required binary placeholder.
  _Fix:_ Validate the capped buffer as UTF-8 and only tolerate an incomplete trailing codepoint caused by the read cap; return \`Prepared::Binary\` for invalid sequences before the boundary.
- ⚪ **[LOW]** \[security\] Changed-only paths can synthesize nodes outside the root — `src/tree.rs:125` · 1/3 models _(security)_
  \`changed\_only\_nodes\` trusts \`changed\_filter\` keys and later uses \`root.join\`, so an absolute or \`..\` path would produce visible nodes outside the tree root. The current Git service should not emit such paths, but this breaks the TreeModel's own root-boundary contract at its public input boundary.
  _Fix:_ Reject or normalize changed-set paths before storing/emitting them, allowing only relative paths with no absolute prefix or parent-dir components.
- ⚪ **[LOW]** \[performance\] visible\_nodes re-walks the filesystem for every expanded dir on every call, with no caching — `src/tree.rs:192` · 1/3 models _(performance)_
  visible\_nodes() rebuilds a WalkBuilder (re-reading ancestor .gitignore files) and readdirs every expanded directory on each call, and it is called repeatedly per keystroke (move\_cursor, selected, clamp\_cursor, plus the presenter). AC-22 only guards launch; interactively expanding many dirs in a large repo means N+1 ignore-aware walks per redraw, with no memoization.
  _Fix:_ Cache the entries of expanded directories and invalidate on expand/collapse/filter/toggle, or have move\_cursor/selected derive length from a cached snapshot instead of recomputing the full walk each time.
- ⚪ **[INFO]** \[correctness\] to\_text fallback leaves SGR parameter text as visible literals — `src/render.rs:327` · 1/3 models _(correctness)_
  If into\_text fails, the fallback is Text::raw(cleaned.replace('\\u{1b}', '')); stripping only the ESC byte from kept SGR sequences leaves e.g. '\[31m' as literal rendered text. Safe (no control byte reaches the terminal) but cosmetically wrong in the unlikely parse-failure path.
  _Fix:_ In the fallback, strip whole SGR sequences (ESC \[ ... m) rather than only the ESC byte, or render the already-stripped plain text without the kept SGR runs.
- ⚪ **[INFO]** \[correctness\] ESC+intermediate sequences (e.g. ESC ( B) drop only 2 bytes, leaving the final byte as literal text — `src/render.rs:370` · 1/3 models _(correctness)_
  The Some(\_) =\> i += 2 branch handles ESC + one byte (like ESC c reset) but mis-handles 3-byte sequences such as ESC ( B (charset designate): it drops ESC and '(', leaving 'B' as visible content. Not a control-sequence bypass (the ESC is gone so no terminal acts on it), just a cosmetic stray character.
  _Fix:_ For ESC followed by an intermediate-range byte (0x20-0x2F), consume the subsequent final byte too before advancing, so no leftover literal leaks.
- ⚪ **[INFO]** \[security\] Root-containment check (canonicalize+starts\_with) does not cover hard links — `src/render.rs:47` · 1/3 models _(security)_
  classify enforces AC-N5 ("no node escapes root") via path.canonicalize().starts\_with(root), which correctly blocks symlink/.. escapes but not hard links: a hard link inside root to an inode outside root canonicalizes to an in-root path and is read and shown. Practical exploitability is low — git clones cannot carry hard links and fs.protected\_hardlinks blocks linking files you cannot already read — so this is an observation against the absolute AC-N5 guarantee, not an exploitable hole in the stated untrusted-repo threat model.
  _Fix:_ Document that containment is path-based (hard links out of scope), or for defense-in-depth refuse regular files whose hard-link count \> 1 when strict containment is required.

smarzban:
## Orchestrator review — PR2 (Tree Model & Content Renderer)

**What this PR implements.** `specs/file-viewer/plan.md` tasks **T-6…T-12**, test-first, repo green between tasks:
- **Tree Model** — gitignore-aware **lazy** enumeration (AC-3/4), show-ignored + changed-only toggles (AC-5/6), per-node status markers (AC-7, tree side), clamped cursor methods, and a perf guard (interactive <1s at 10k files, AC-22). Changed-only mode is built from the changed-set itself, so **deleted files — and files under a wholly-deleted directory — remain reviewable**.
- **Content Renderer** — bounded classify (binary placeholder AC-12; ≥1 MB/5000-line preview + notice AC-13), escape-sequence neutralization (AC-27), and delegation to injected markdown/diff/syntax renderers with a plain-text fallback + capability notice (AC-8/9/10/24/25), plus an in-process perf guard (<300ms at 1 MB, AC-23 in-process bound).

~60 tests pass; release binary builds; output warning-clean.

**Review journey (5 rounds).** This was reviewed across **five panel rounds** (holistic ×N + `lens-spec` + `lens-security`). The Content Renderer and Tree Model were driven to a genuinely hardened state — **every real gating finding was fixed**, including several regressions the panels caught in my own earlier fixes:
- **Untrusted file bytes**: bounded reads (no slurp); binary files emit no raw bytes; **all C0 + C1 control sequences neutralized** (AC-27) — including the ANSI-parse error fallback path; `classify(root, path)` refuses to read a symlink/`..` escaping the root or a FIFO/device (AC-N5, no leak/hang).
- **Untrusted file names**: the `{name}` language hint is sanitized to a safe basename (no shell/option injection), preserving the secure stdin design while enabling `bat` highlighting (AC-10).
- **External renderers**: time-bounded (kill + fallback), stdout size-capped (16 MB), bounded post-EOF wait.
- **Correctness**: base-hint threading already done in PR1; deleted-file diffs render in Diff mode; oversized diffs are preview-capped; cursor stays in bounds across filter changes.

**Dismissed earlier / now advisory.** The classify **TOCTOU** (canonicalize→open race) is an accepted INFO/LOW: the viewer runs **read-only as the user**, so a sub-ms race could only read files the user can already read (no privilege boundary) — consistent with PR1's stance.

**Deferred advisories (LOW/INFO, non-blocking follow-ups, in the gate comment).** `read_to_end` failure → empty pane (vs placeholder); `visible_nodes` re-walk caching; the up-to-2× timeout bound; bidi/invisible Unicode neutralization; hard-link containment; large-invalid-UTF-8-without-NUL classification; and a few test-coverage gaps (OSC/C1 paths, 16 MB cap). Reasonable to pick up in a later PR.

**What this PR does NOT cover (by design — later PRs).** Input Dispatcher / Presenter / Editor Launcher / Host Adapter runtime (PR3), Session Controller + run loop incl. off-thread rendering (PR4), e2e + docs (PR5).

**Decision: ✅ Approve.** No blocking findings (0 critical/high/medium); the Tree Model and Content Renderer are correct, well-tested, and hardened for the untrusted-content threat model.

PR #3, opened 2026-06-20
PR3: Input Dispatcher, Presenter, Editor Launcher, Host Adapter (T-13…T-17)

smarzban:
## What this PR implements

PR3 of the layer-based build — the **interaction & boundary layer** (plan tasks T-13–T-17), each built test-first (red→green) on top of the PR1/PR2 foundation. The repo stays green between tasks; 101 tests pass.

| Task | Component | Files | Advances |
| --- | --- | --- | --- |
| **T-13** | Input Dispatcher | `src/intent.rs`, `src/input.rs` | AC-18, AC-N3 |
| **T-14** | Presenter (two-column) | `src/presenter.rs` | AC-3, AC-7, AC-13, AC-25 (display) |
| **T-15** | Presenter (narrow-split) | `src/presenter.rs` | AC-21 |
| **T-16** | Editor Launcher | `src/editor.rs` | AC-19, AC-N1 |
| **T-17** | Host Adapter | `src/host.rs` | AC-19 (new-pane), AC-26 |

### Highlights
- **Input Dispatcher** — closed `Intent` enum (11 variants) + `map_key`; keyboard-complete (AC-18) proven via a bindings table covering `Intent::ALL`; char bindings gated to no-modifier so chords like Ctrl+C stay clear. The no-edit invariant (AC-N3) is enforced by an exhaustive match that fails to compile if an edit intent is ever added.
- **Presenter** — ratatui two-column layout: bordered file tree (recursive indentation AC-3, M/A/D/? status markers AC-7) + content pane with a bounded notices strip (truncation AC-13, renderer-fallback AC-25). Under 80 cols it drops to a single focused full-width column (AC-21). Behavioral asserts + `insta` snapshots.
- **Editor Launcher** — pure hand-off (`Target::Editor` / `Target::NewPane`) over an injected `Spawner`; no file I/O (AC-N1); launch failure is an `Err`, never a panic.
- **Host Adapter** — defensive parse of `HERDR_PLUGIN_CONTEXT_JSON` (malformed/missing → minimal `{ cwd }`, AC-26); `open_pane` issues the `pane split → parse result.pane.pane_id → pane run` sequence over an injected `HerdrCli` seam.

### Security (untrusted boundary, written upfront)
The Host Adapter shells out to herdr with an untrusted file name:
- file name **POSIX single-quoted** into the run command (command-injection guard; round-trip proven),
- relative path beginning with `-` is **`./`-prefixed** (editor option-injection guard),
- both the current and split-returned **pane ids validated** as safe argv tokens (CLI option-injection guard); a garbled/flag-like split result is an `Err`, not a second call.

### Out of scope / deferred
- Wiring it all together (Session Controller, run loop) — **PR4 (T-18–T-20)**.
- e2e pty tests + README/manual-verification — **PR5 (T-21–T-25)**.
- The `pane list` fallback for current-pane-id (used only when context lacks it) is left to PR4 wiring — its CLI output schema isn't pinned in techstack, so it's resolved at integration rather than guessed here.

Verification: `cargo test` → 101 pass / 0 fail; `cargo build --release` clean; `cargo clippy` clean on all new modules.


smarzban:
## Review Gate — Round 1

Findings: 12 total — 0 critical · 0 high · 5 medium · 7 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-spec + tools · models: ollama:kimi-k2.7-code:cloud, ollama:glm-5.2:cloud, claude:claude-opus-4-8, codex:gpt-5.5, deterministic

### Must fix
- 🟠 **[MEDIUM]** \[correctness\] EDITOR values with arguments cannot launch — `src/editor.rs:47` · 2/4 models _(correctness)_
  EditorLauncher documents accepting an editor command from $EDITOR, but Target::Editor passes the entire value as argv\[0\]. Common valid configs like "code --wait" or "vim -f" are treated as executable names and fail, so open-in-editor does not reliably satisfy AC-19.
  _Fix:_ Represent the configured editor as program plus arguments, append the selected file as the final argv element, and update the new-pane path to quote/render the same argv parts consistently.
- 🟠 **[MEDIUM]** \[security\] Tree labels and content titles are not escape-sanitized — `src/presenter.rs:56` · 2/4 models _(security)_
  node\_name renders raw file-name bytes into tree rows and the content title, so a repository-controlled filename containing ESC/CSI or backspace bytes can move the cursor, clear the screen, or spoof status markers in the viewer pane.
  _Fix:_ Sanitize node\_name output before turning it into a Line or title by stripping ESC sequences and C0 control bytes (e.g., a small dedicated sanitizer, stricter than the content SGR-preserving one).
- 🟠 **[MEDIUM]** \[security\] unsanitized file names can emit terminal control sequences — `src/presenter.rs:76` · 1/4 models _(security)_
  Repo-controlled file names are converted with \`to\_string\_lossy()\` and rendered directly into the tree rows and selected-file title, bypassing the terminal-control stripping used for file contents. A malicious filename containing ESC/OSC/C0 bytes can manipulate the terminal display or clipboard when the viewer draws it.
  _Fix:_ Sanitize every presenter string that crosses from paths/notices into ratatui output, using the same control-sequence stripping policy as content rendering, and add regression tests with filenames containing ESC/OSC sequences.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[correctness\] \`shell\_quote\_path\` uses \`to\_string\_lossy\`, mangling non-UTF-8 filenames — `src/host.rs:141` · 1/4 models _(correctness)_
  \`file.to\_string\_lossy()\` replaces invalid-UTF-8 bytes with U+FFFD, so the editor is handed a different filename than the one on disk for such paths — the hand-off silently opens the wrong file. This diverges from the Git Service, which deliberately preserves raw bytes via \`OsStrExt\`.
  _Fix:_ Quote the OsStr directly (operate on bytes/chars without lossy conversion) so the editor receives the exact filename; only the flag-safe \`./\` prefix needs the string form, and that can be decided via \`as\_bytes().first()\`.
- ⚪ **[LOW]** \[correctness\] Editor hand-off corrupts non-UTF-8 file paths via to\_string\_lossy — `src/host.rs:214` · 1/4 models _(correctness)_
  git.rs and tree.rs deliberately preserve raw filename bytes (OsStrExt) so non-UTF-8 names map to real paths, but open\_pane builds the \`pane run\` command with file.to\_string\_lossy(), replacing invalid bytes with U+FFFD. Pressing 'e' on a non-UTF-8-named file would open the editor on a non-existent/garbled path.
  _Fix:_ Document the limitation explicitly, or preserve bytes where the herdr CLI protocol permits; at minimum surface a notice when the path is not valid UTF-8 rather than silently handing the editor a corrupted name.
- ⚪ **[LOW]** \[maintainability\] \`from\_env\` parses \`pane\_id\` from context then discards it — `src/host.rs:55` · 1/4 models _(maintainability)_
  \`RawContext\` has \`pane\_id\` and \`from\_env\`/\`parse\_context\` deserialize it, but \`LaunchContext\` has no such field, so the value is dropped. Any caller needing the current pane id (required by \`open\_pane\` for \`Target::NewPane\`) must re-read and re-parse \`HERDR\_PLUGIN\_CONTEXT\_JSON\` via \`current\_pane\_id\`, coupling the controller to the env layout and parsing twice.
  _Fix:_ Either add \`pane\_id: Option\<String\>\` to \`LaunchContext\` (validated once at parse time) and have \`current\_pane\_id\` derive from it, or have \`from\_env\` return both the context and an optional pane id so the JSON is parsed once.
- ⚪ **[LOW]** \[correctness\] Key release events can trigger intents — `src/input.rs:15` · 1/4 models _(correctness)_
  map\_key ignores KeyEventKind, so crossterm release or repeat events from enhanced keyboard modes are mapped the same as key presses. This can make toggles fire on release/repeat and leave the UI in the wrong state.
  _Fix:_ Return None for non-press events, or explicitly decide which intents may accept KeyEventKind::Repeat while rejecting Release.
- ⚪ **[LOW]** \[maintainability\] Intent::ALL is not verified to cover every enum variant — `src/intent.rs:38` · 1/4 models _(maintainability)_
  The ALL array is hand-curated; if a future Intent variant is added to the enum but omitted from ALL, the keyboard-completeness test will silently pass while AC-18 is violated.
  _Fix:_ Add a test asserting Intent::ALL.len() == std::mem::variant\_count::\<Intent\>() so the list cannot drift from the actual enum.
- ⚪ **[LOW]** \[correctness\] Presenter decides wide/narrow from \`state.width\` but draws from \`frame.area()\` — `src/presenter.rs:157` · 1/4 models _(correctness)_
  \`draw\` branches on \`state.width \< NARROW\_SPLIT\` but takes all geometry from \`frame.area()\`. If the controller ever feeds a stale \`state.width\` that disagrees with the live frame (resize between observe and draw), the viewer will render two columns into a narrow frame or hide a column in a wide one. The doc comment asserts they are equal, but that invariant is unenforced.
  _Fix:_ Drive the split decision off \`frame.area().width\` (clamping \`state.width\` to it) or assert/clip \`state.width\` to the frame width before the branch, so the layout and geometry can't disagree.
- ⚪ **[LOW]** \[security\] Untrusted filenames rendered in tree/titles bypass the AC-27 escape-neutralizer — `src/presenter.rs:521` · 1/4 models _(security)_
  File contents are explicitly sanitized via render::to\_text, but repo-controlled filenames (which on Unix may contain ESC/C0 control bytes) are rendered straight from to\_string\_lossy in tree\_row/node\_name/block titles. Safety here relies implicitly on ratatui dropping zero-width control graphemes rather than the explicit neutralizer used for content — a defense-in-depth gap given the stated untrusted-repo threat model; if that ratatui behavior does not hold, a crafted filename could inject terminal control sequences.
  _Fix:_ Route displayed filenames through the same control/escape stripping used for content (e.g. a shared sanitizer), or add a presenter test feeding a control-char filename and asserting the bytes never reach the cell buffer.

- **[MEDIUM]** \[correctness\] Editor path is not shell-escaped in the herdr pane run command — `src/host.rs:117`
  _Dismissed:_ Read src/host.rs open\_pane: command = format!("{} {}", editor.to\_string\_lossy(), shell\_quote\_path(file)). The untrusted, repo-controlled input is the FILE name, and it IS POSIX single-quoted by shell\_quote\_path (round-trip proven in tests/host\_context.rs). The editor string is application configuration ($EDITOR / user config), not repo/attacker-controlled, and is deliberately left unquoted so editors-with-flags (e.g. 'code --wait') work — this mirrors git's own GIT\_EDITOR convention (editor = trusted shell text, file = escaped data). No repo-controlled value reaches an unquoted shell position, so there is no injection vector under the threat model. Adding a doc comment to make the trust boundary explicit.
- **[MEDIUM]** \[correctness\] Multi-line notices are clipped as one row — `src/presenter.rs:127`
  _Dismissed:_ Checked the notice sources: notices are produced internally by the Content Renderer / Git Service as fixed single-line strings ('Showing first 5000 lines (truncated)', 'delta not found — showing plain diff'); none contain a newline, so the one-Line-per-notice rendering in draw\_content is correct. Single-model, contested; advisory by gating calibration. Logged as a follow-up to revisit if multi-line notices are ever introduced.

smarzban:

**What this PR implements:** the interaction & boundary layer (T-13…T-17) — Input Dispatcher (`map_key`/`Intent`), Presenter (two-column + narrow-split, status markers, notices), Editor Launcher (hand-off), Host Adapter (context parse + herdr open-pane). 101 tests, all test-first.

**Round-1 panel:** holistic ×4 (kimi-k2.7-code, glm-5.2, opus-4-8, gpt-5.5) + lens-security (opus, gpt-5.5) + lens-spec (glm, kimi) + deterministic scan. Full panel, no model lost (deterministic secret-scan skipped — gitleaks not installed; the diff is Rust source + tests, no secret surface).

### Blocking — what I'm fixing
1. **AC-27 / displayed file names (`presenter.rs`, 4 models).** `node_name` renders raw repo-controlled file names into tree rows and the content-pane title without the control-stripping that file *content* gets via `render::to_text`. I verified empirically that ratatui *already* strips control characters (a file named `evil\x1b[2J\x1b[10;10Hpwned` produced **zero** control codepoints in the rendered buffer — `has_esc=false`), so there is **no live AC-27 violation**. But AC-27 is a security criterion and that guarantee must live in *our* code, not rest on a ratatui internal — **fixing** by routing every displayed label/title through an explicit control-sanitizer + a regression test feeding an ESC-laden filename.
2. **`$EDITOR` with arguments (`editor.rs:47`, 2 models).** `Target::Editor` passes the whole editor string as `argv[0]`, so `"code --wait"` can't launch — a real AC-19 gap. **Fixing** by splitting the editor into program + args before appending the file.

### Dismissed (code-checked)
- **Editor unquoted in the `pane run` command (`host.rs:117`, medium ×2).** The untrusted, repo-controlled input is the *file name*, which **is** single-quoted (round-trip proven). The editor is application config (`$EDITOR`), not repo/attacker-controlled, and must stay unquoted to support editors-with-flags — this mirrors git's own `GIT_EDITOR` handling (editor = trusted shell text, file = escaped data). No repo-controlled value reaches an unquoted position. Adding a doc comment to make the trust boundary explicit.
- **Multi-line notices clipped (`presenter.rs:127`, single model).** Notices are internally-produced single-line strings; none contain newlines. Advisory.

### Advisory — folding in the cheap correctness ones, noting the rest
- Will also **fix** the low `state.width` vs `frame.area()` split-decision divergence (`presenter.rs:157`) by driving the decision off the live frame width.
- **Deferring:** non-UTF-8 paths via `to_string_lossy` in the herdr hand-off (rare on the target platforms; will document the limitation), key-release-event filtering (belongs in the T-20 run loop), and the `Intent::ALL` drift-guard / `pane_id` parse-once tidy-ups (logged as follow-ups).

**Decision: 🔴 Request changes** — fixing the blocking items, then a verification round.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ tools + verify · models: codex:gpt-5.5, deterministic

### Progress since Round 1
✅ Resolved (3): \[correctness\] EDITOR values with arguments cannot launch; \[security\] Tree labels and content titles are not escape-sanitized; \[security\] unsanitized file names can emit terminal control sequences — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

**Verification round.** Single-model verifier (codex `gpt-5.5`) scoped to the Round-1 findings + the fix diff (`7619c06...fa93cd5`), plus the deterministic scan. The first verifier (opus) returned an unparseable non-vote — I did **not** treat that as a pass; I re-ran with a reliable parser, which returned `[]` (clean). All three Round-1 blocking findings are resolved at HEAD with **no regression**; scan clean.

### What this PR implements
The interaction & boundary layer (T-13…T-17): Input Dispatcher (`map_key`/closed `Intent` set, keyboard-complete, no edit intent), Presenter (two-column + narrow-split focus-toggle, recursive tree, status markers, notices), Editor Launcher (hand-off to editor / new pane), Host Adapter (defensive context parse + hardened herdr open-pane). 105 tests, all test-first; `cargo build --release` + clippy clean.

- **AC-27 (security):** displayed file names + content title now routed through `sanitize_label` (drops C0/DEL/C1 control chars). Verified empirically that ratatui already strips controls (no live vuln existed), but the guarantee now lives in our code, with an end-to-end regression test.
- **AC-19:** `$EDITOR`-with-flags (e.g. `code --wait`) now launches correctly (program + args split).
- **Robustness:** narrow/wide split decision driven off the live frame width, so layout can't desync from a stale `state.width`.
- Doc: explicit editor(trusted shell)/file(escaped) trust boundary in the Host Adapter.

### Dismissed (code-checked, Round 1)
- Editor unquoted in `pane run` — the untrusted file is single-quoted; the editor is trusted config (git `GIT_EDITOR` convention).
- Multi-line notices clipped — notices are internally-produced single-line strings.

### Deferred / not covered (follow-ups, non-blocking)
- Non-UTF-8 file paths in the herdr hand-off use `to_string_lossy` (rare on linux/macos; documented limitation).
- Key-release-event filtering belongs in the T-20 run loop (PR4).
- Minor tidy-ups: `Intent::ALL` drift-guard, parse-`pane_id`-once. Logged, low.
- Wiring (Session Controller, run loop), e2e/pty tests, README — **PR4/PR5**.

**Decision: ✅ Approve** (agrees with the `pass` verdict).

> ⏸️ **Not merging.** Per the PR owner's request, this PR is left open for their review before landing — the gate has passed but the merge is intentionally deferred to them.

PR #4, opened 2026-06-20
PR4: Session Controller, off-thread rendering, main wiring (T-18…T-20)

smarzban:
## PR4 — Session Controller, off-thread rendering, main wiring (T-18…T-20)

Layer 4 of the plan: the interaction orchestrator and the live run loop that ties
all prior components together into a working viewer.

### Tasks
- **T-18 — Session Controller** (`src/controller.rs`): holds the ephemeral session
  state (root, git-presence, baseline, filter flags, per-file view overrides, cursor,
  focus, observed width) and turns each `Intent` into state changes + `Effects`
  (`redraw`/`quit`/`clear`). Git Service, Content Renderer, and Editor Launcher are
  behind injected traits (`GitService`/`ContentProvider`/`EditorHandoff`) so the
  controller is unit-testable with stubs; the Tree Model is driven directly.
  - Git-only intents are inert without a repo (AC-26); a component failure becomes a
    non-fatal notice, never a crash; no intent mutates the filesystem (AC-N1/AC-N3).
  - Covers AC-5, AC-6, AC-11, AC-16, AC-20 (trigger), AC-21 (trigger), AC-26.
- **T-19 — off-thread rendering** (AC-23): the Content Renderer runs on a worker
  thread; selecting a file *dispatches* a render job and `handle()` returns
  immediately, so input never blocks on a slow external renderer. `poll()` drains
  finished renders, applying only the latest selection's (stale results dropped by a
  monotonic sequence).
- **T-20 — main wiring** (`src/app.rs`, `src/lib.rs::run()`): reads the herdr context →
  resolves root/git-presence → builds the live components → drives a draw→input→poll
  loop over a ratatui terminal (crossterm raw mode + restore, incl. ratatui's panic
  hook). Editor hand-off suspends/restores the TUI around a blocking `$EDITOR` spawn.
  Covers AC-17 (launch behavior) and AC-20.

### Tests (test-first, red→green)
- `tests/controller.rs` (10) — intent orchestration, inert-without-git, editor-error
  notice, read-only invariant over the whole intent vocabulary.
- `tests/controller_async.rs` (2) — handle() doesn't block on a slow render; a
  superseded render never clobbers a newer selection.
- `tests/cli_smoke.rs` (1, expectrl pty) — the built binary draws the tree and exits 0
  on the close key.
- Full suite: **118 passing**, 0 failing. New modules clippy-clean; release builds.

### Notes
- First commit adds `.review-gate/` to `.gitignore` (gate run logs stay local).
- The `presenter::ViewState::width` field (kept in PR3) is now read by
  `Controller::view_state()`.
- Pre-existing clippy warnings in `src/git.rs` / `tests/git_status.rs` predate this PR
  and are outside its diff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@
 
 # Local Claude Code tooling (skills, local settings) — not part of the plugin
 .claude/
+
+# Review-gate run logs / dismissal records — local audit trail, not part of the plugin
+.review-gate/
--- a/src/app.rs
+++ b/src/app.rs
@@ -26,6 +26,7 @@ use std::ffi::{OsStr, OsString};
 use std::io;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::sync::Arc;
 use std::time::Duration;
 
 /// How long the input poll blocks each tick before draining finished off-thread renders, so
@@ -41,7 +42,7 @@ pub fn run() -> io::Result<()> {
     let resolved = root::resolve(&ctx);
     let baseline = git::default_baseline(&resolved);
 
-    let git: Box<dyn GitService> = Box::new(LiveGit {
+    let git: Arc<dyn GitService> = Arc::new(LiveGit {
         // In a non-repo there is no repo_root; git is never queried then, but a path is still
         // required, so fall back to the tree root.
         repo_root: resolved.repo_root.clone().unwrap_or_else(|| resolved.root.clone()),
@@ -65,14 +66,20 @@ pub fn run() -> io::Result<()> {
     outcome
 }
 
-/// Draw, read one input (or time out), drain renders; repeat until the Close intent.
+/// Draw (only when something changed), read one input (or time out), drain renders; repeat
+/// until the Close intent. Drawing only when `dirty` avoids re-walking the filesystem (the
+/// tree enumeration in `view_state`) on every idle tick.
 fn event_loop(terminal: &mut DefaultTerminal, controller: &mut Controller) -> io::Result<()> {
+    let mut dirty = true; // paint the first frame
     loop {
-        terminal.draw(|frame| {
-            controller.set_width(frame.area().width);
-            let view: ViewState = controller.view_state();
-            presenter::draw(frame, &view);
-        })?;
+        if dirty {
+            terminal.draw(|frame| {
+                controller.set_width(frame.area().width);
+                let view: ViewState = controller.view_state();
+                presenter::draw(frame, &view);
+            })?;
+            dirty = false;
+        }
 
         if event::poll(TICK)?
             && let Event::Key(key) = event::read()?
@@ -82,13 +89,17 @@ fn event_loop(terminal: &mut DefaultTerminal, controller: &mut Controller) -> io
             let fx = controller.handle(intent);
             if fx.clear {
                 terminal.clear()?; // an editor took the screen — force a full repaint
+                dirty = true;
             }
             if fx.quit {
                 return Ok(());
             }
+            dirty |= fx.redraw;
         }
-        // Surface any content finished by the render worker on the next draw (AC-23).
-        controller.poll();
+        // A render finished by the worker becomes visible on the next draw (AC-23).
+        if let Some(fx) = controller.poll() {
+            dirty |= fx.redraw;
+        }
     }
 }
 
@@ -142,12 +153,22 @@ struct LiveEditor {
 impl EditorHandoff for LiveEditor {
     fn open(&mut self, file: &Path) -> io::Result<bool> {
         let Some(editor) = self.editor.clone() else {
+            // No terminal change yet, so nothing to restore.
             return Err(io::Error::other("no editor configured (set $EDITOR)"));
         };
-        suspend_tui()?;
-        let launched = EditorLauncher::new(editor).open(file, Target::Editor, &mut ProcessSpawner);
+        // Once we touch the terminal we must always try to restore it, even if suspending or
+        // the editor itself fails — otherwise the viewer would keep running with raw mode off
+        // or outside the alternate screen.
+        let suspended = suspend_tui();
+        let launched = if suspended.is_ok() {
+            EditorLauncher::new(editor).open(file, Target::Editor, &mut ProcessSpawner)
+        } else {
+            // Don't run the editor over a half-suspended terminal.
+            Ok(())
+        };
         let resumed = resume_tui();
-        // Report the editor's own failure first, then any failure to restore the terminal.
+        // Report the first failure in order: suspend, then the editor, then restore.
+        suspended?;
         launched?;
         resumed?;
         Ok(true) // the editor drew over the screen → the run loop forces a full repaint
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,0 +1,210 @@
+//! App wiring — assemble the real components and run the terminal event loop (T-20).
+//!
+//! [`run`] is the binary's body: read the herdr launch context, resolve the root and
+//! git-presence, build the live Git Service / Content Renderer / Editor Launcher behind the
+//! controller's traits, then drive a draw → input → poll loop over a ratatui terminal until
+//! the Close intent (AC-20). The terminal is restored on every exit path — including a panic,
+//! via the hook `ratatui::try_init` installs.
+
+use crate::controller::{
+    Components, ContentProvider, Controller, EditorHandoff, GitService, RenderResult,
+};
+use crate::editor::{EditorLauncher, Spawner, Target};
+use crate::git::{self, Baseline, Status};
+use crate::presenter::{self, ViewState};
+use crate::render::{self, Prepared, Renderers};
+use crate::view_policy::ViewMode;
+use crate::{host, input, root};
+use crossterm::event::{self, Event, KeyEventKind};
+use crossterm::execute;
+use crossterm::terminal::{
+    EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode,
+};
+use ratatui::DefaultTerminal;
+use std::collections::BTreeMap;
+use std::ffi::{OsStr, OsString};
+use std::io;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::time::Duration;
+
+/// How long the input poll blocks each tick before draining finished off-thread renders, so
+/// late content appears promptly without the loop busy-spinning.
+const TICK: Duration = Duration::from_millis(50);
+
+/// Per-render wall-clock budget for an external renderer before the plain-text fallback.
+const RENDER_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Wire the components and run the viewer until the user closes it.
+pub fn run() -> io::Result<()> {
+    let ctx = host::from_env();
+    let resolved = root::resolve(&ctx);
+    let baseline = git::default_baseline(&resolved);
+
+    let git: Box<dyn GitService> = Box::new(LiveGit {
+        // In a non-repo there is no repo_root; git is never queried then, but a path is still
+        // required, so fall back to the tree root.
+        repo_root: resolved.repo_root.clone().unwrap_or_else(|| resolved.root.clone()),
+        base_hint: resolved.base_branch.clone(),
+    });
+    let content: Box<dyn ContentProvider> =
+        Box::new(LiveContent { root: resolved.root.clone(), renderers: default_renderers() });
+    let editor: Box<dyn EditorHandoff> =
+        Box::new(LiveEditor { editor: std::env::var_os("EDITOR") });
+
+    let mut controller = Controller::new(
+        resolved.root.clone(),
+        resolved.is_git_repo,
+        baseline,
+        Components { git, content, editor },
+    );
+
+    let mut terminal = ratatui::try_init()?;
+    let outcome = event_loop(&mut terminal, &mut controller);
+    ratatui::try_restore()?;
+    outcome
+}
+
+/// Draw, read one input (or time out), drain renders; repeat until the Close intent.
+fn event_loop(terminal: &mut DefaultTerminal, controller: &mut Controller) -> io::Result<()> {
+    loop {
+        terminal.draw(|frame| {
+            controller.set_width(frame.area().width);
+            let view: ViewState = controller.view_state();
+            presenter::draw(frame, &view);
+        })?;
+
+        if event::poll(TICK)?
+            && let Event::Key(key) = event::read()?
+            && key.kind == KeyEventKind::Press
+            && let Some(intent) = input::map_key(key)
+        {
+            let fx = controller.handle(intent);
+            if fx.clear {
+                terminal.clear()?; // an editor took the screen — force a full repaint
+            }
+            if fx.quit {
+                return Ok(());
+            }
+        }
+        // Surface any content finished by the render worker on the next draw (AC-23).
+        controller.poll();
+    }
+}
+
+/// The live Git Service: read-only queries against the resolved repository (AC-7/9/16).
+struct LiveGit {
+    repo_root: PathBuf,
+    base_hint: Option<String>,
+}
+
+impl GitService for LiveGit {
+    fn status(&self) -> BTreeMap<PathBuf, Status> {
+        git::status(&self.repo_root)
+    }
+    fn changed_set(&self, baseline: Baseline) -> BTreeMap<PathBuf, Status> {
+        git::changed_set(&self.repo_root, baseline, self.base_hint.as_deref())
+    }
+    fn diff(&self, rel_path: &Path, baseline: Baseline) -> String {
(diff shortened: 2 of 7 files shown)


smarzban:
## Review Gate — Round 1
🚫 **BLOCK** — 5 blocking finding(s) must be resolved or justified.

Findings: 12 total — 0 critical · 0 high · 6 medium · 5 low · 1 info.
_Reviewed by:_ holistic-glm2 + holistic-gpt + holistic-kimi + holistic-opus + security-gpt + subtle-gpt + subtle-kimi + tools · models: ollama:kimi-k2.7-code:cloud, ollama:glm-5.2:cloud, claude:claude-opus-4-8, codex:gpt-5.5, deterministic

### Must fix
- 🟠 **[MEDIUM]** \[correctness\] editor hand-off error path leaves the screen corrupted (no forced clear) — `src/app.rs:151` · 1/4 models _(correctness)_
  LiveEditor::open suspends (LeaveAlternateScreen + disable\_raw\_mode) and resumes (EnterAlternateScreen) the terminal, then returns Err when the editor binary is missing or exits non-zero. The controller's Err branch returns Effects::redraw() WITHOUT clear, but the success path returns clear:true precisely because re-entering the alternate screen blanks it while ratatui's diff cache is stale — so on failure the tree/content are not repainted and the display is left corrupted until the next full redraw.
  _Fix:_ Force a full clear whenever the hand-off toggled the terminal, even on failure — e.g. have the controller set clear:true on the OpenInEditor Err branch, or change EditorHandoff so an error that occurred after suspend still signals 'terminal taken' so the run loop clears.
- 🟠 **[MEDIUM]** \[concurrency\] FIFO render worker lets stale jobs block the current selection — `src/controller.rs:143` · 2/4 models _(concurrency)_
  Interleaving: a slow render for file A starts, the user rapidly moves through B/C/D, and the single worker must finish every obsolete queued job before rendering D. The sequence check drops stale results only after they finish, so the selected file can show stale or blank content for N \* renderer\_timeout and the unbounded channel can grow under repeated navigation.
  _Fix:_ Coalesce rendering to the latest request: use a bounded/latest-only channel, drain pending jobs before rendering, or add a shared generation check that skips obsolete jobs before invoking the renderer.
- 🟠 **[MEDIUM]** \[correctness\] toggle\_ignore leaves content pane stale when selection shifts — `src/controller.rs:292` · 2/4 models _(correctness)_
  Toggling show\_ignored rebuilds the visible tree and clamps the cursor, but does not dispatch a render; if ignored files are inserted before the cursor, the cursor now points to a different file while the content pane still shows the previous selection.
  _Fix:_ Call dispatch\_render() at the end of toggle\_ignore so the content pane matches the newly selected node, matching the pattern used by toggle\_changed\_only.
- 🟠 **[MEDIUM]** \[caching\] Git state cache is not invalidated after returning from the editor — `src/controller.rs:343` · 1/4 models _(caching)_
  After \`OpenInEditor\` returns, the editor may have modified the selected file, but the cached status/changed-set and current content are left unchanged. The tree markers, changed-only filter, and default diff-vs-content mode can silently reflect the pre-edit repository state until some unrelated action or restart.
  _Fix:_ On successful editor return, refresh \`git.status()\` and \`git.changed\_set(self.baseline)\`, update the tree status/filter caches, and dispatch a render for the current selection before redrawing.
- 🟠 **[MEDIUM]** \[performance\] git diff is read unbounded on the input thread before the size cap — `src/controller.rs:378` · 4/4 models _(performance)_
  dispatch\_render synchronously runs git diff and buffers the entire output into a String on the input thread; the size cap in render::cap\_preview only runs later in the worker, so a large diff blocks navigation and can spike memory before any cap applies, contradicting the non-blocking intent of AC-23.
  _Fix:_ Move the diff query into the render worker or cap it synchronously on the input thread so diff-mode content stays bounded and non-blocking.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[error-handling\] partial suspend\_tui failure leaves the viewer running with raw mode off — `src/app.rs:181` · 1/4 models _(error-handling)_
  suspend\_tui does disable\_raw\_mode()? then LeaveAlternateScreen. If the first succeeds but the second fails, open() returns Err (surfaced as a notice) but raw mode is now disabled while the event loop keeps running, leaving the terminal in a degraded state until Close calls try\_restore().
  _Fix:_ On suspend\_tui partial failure, re-enable raw mode before returning the error so the loop resumes in a consistent state, or treat any suspend failure as fatal and restore+exit.
- ⚪ **[LOW]** \[performance\] event loop redraws unconditionally every tick, re-walking the filesystem while idle — `src/app.rs:71` · 1/4 models _(performance)_
  terminal.draw runs at the top of every loop iteration (≤50ms), and view\_state() calls tree.visible\_nodes() which freshly walks every expanded directory via the ignore crate and clones the full content Text (twice, also in draw\_content) — so the viewer does \~20 filesystem walks/sec plus full-content clones continuously even when the user is just reading. The Effects.redraw flag and poll()'s returned effect (intended to gate redraws) are ignored by the loop.
  _Fix:_ Only redraw when something changed: track a dirty flag from controller.handle()'s Effects.redraw and poll()'s return, and skip terminal.draw when neither fired (the TICK still drains the render channel).
- ⚪ **[LOW]** \[maintainability\] poll's redraw effect is discarded in the live event loop — `src/app.rs:91` · 2/4 models _(maintainability)_
  Controller::poll returns Option\<Effects\> whose redraw flag is thrown away; the event loop redraws unconditionally every tick, so the Effects::redraw contract is partly dead code and misleading for future callers.
  _Fix:_ Either use poll's return value to decide whether to redraw, or change poll to return a simpler bool/None so the API contract matches actual usage.
- ⚪ **[LOW]** \[performance\] unbounded job channel carries uncapped raw\_diff strings — `src/controller.rs:141` · 1/4 models _(performance)_
  job\_tx is an unbounded mpsc channel and each RenderJob for diff mode owns the full raw\_diff String (uncapped by git::diff). Rapid navigation (holding j) over large-diff files queues many jobs, each holding a full diff, so memory can spike before the worker drains them.
  _Fix:_ Bound the job channel (e.g. capacity 1, or drain-and-replace on send) so only the latest dispatch is retained; stale queued jobs are useless anyway since poll drops non-latest results.
- ⚪ **[LOW]** \[time\] Async test uses a tight elapsed-time threshold — `tests/controller\_async.rs:83` · 1/4 models _(time)_
  The test asserts \`handle()\` completes in less than half of a 150 ms artificial delay, which can fail under scheduler stalls even when the implementation is correctly asynchronous. That makes the test nondeterministic on loaded CI hosts.
  _Fix:_ Make the test deterministic by synchronizing with the renderer stub through a channel, or use a much looser upper bound that only catches accidental blocking for the full render duration.
- ⚪ **[INFO]** \[correctness\] global panic hook (try\_init) would restore the terminal from the worker thread on a render panic — `src/app.rs:62` · 1/4 models _(concurrency)_
  ratatui::try\_install registers a global panic hook calling try\_restore(). A panic in the off-thread render worker (controller.rs:143) would invoke that hook on the worker thread, disabling raw mode / leaving the alt screen while the main loop is still drawing — garbling the terminal. Current external renderers exit rather than panic, so this is latent.
  _Fix:_ Catch panics in the worker (std::panic::catch\_unwind around content.render) and convert to a fallback notice instead of letting the global hook run on the worker thread.

- **[MEDIUM]** \[time\] renderer timeout budget is spent twice — `src/render.rs:286`
  _Dismissed:_ Out of this PR's diff: src/render.rs::run\_renderer is unchanged PR2 code (git diff origin/main...HEAD touches controller.rs/app.rs/lib.rs and tests, not render.rs). The doubled budget (recv\_timeout for stdout, then wait\_bounded for exit) is a bounded worst-case on an already-degrading renderer path that still terminates and falls back to plain text + a notice. Logged as a follow-up for a render.rs revision; not introduced or worsened by PR4.

smarzban:

**Decision: 🔴 Request changes** (agrees with the gate's BLOCK).

I ran the full round-1 panel — holistic ×4 (kimi, glm, opus, gpt-5.5; glm's first run
errored on the backend and was re-run to restore the 4th lineage), plus two warranted
lenses (`lens-subtle-correctness` ×2 for the new worker-thread/channel concurrency, and
`lens-security` ×1 for the `$EDITOR` subprocess + terminal-state handling), plus the
deterministic scan (clean; gitleaks not installed → secret scan skipped). No critical/high;
all findings medium and below. I read the code behind every gating cluster.

**Lens decision (written out):** tests — skip (3 new test files cover the surface; holistic
×4 covers quality). spec — skip (the advanced ACs are mechanical; holistic + the plan's AC
map cover them). security — **fired** (subprocess spawn + raw-mode toggling). privacy /
contracts / migrations — skip (none apply). subtle-correctness — **fired** (worker thread +
mpsc + stale-result ordering + suspend/restore). simplify — skip in the gate (advisory; run
as a process, not a panel).

**Fixing (this round's blockers + the corroborating advisories):**
- **git diff on the input thread** (`controller.rs:378`, 4/4 models) — the single strongest
  finding. `GitService` is now `Send + Sync` and `Arc`-shared; the **render worker fetches
  the diff**, so nothing git-related (and no diff text through the channel — resolves the
  "uncapped raw_diff in channel" advisory too) touches the input thread. True AC-23 non-blocking.
- **FIFO worker starves the current selection** (`controller.rs:143`, 2/4) — the worker now
  **collapses its backlog** (drains to the latest queued job) so rapid navigation never waits
  behind superseded renders.
- **toggle_ignore leaves the pane stale** (`controller.rs:292`, 2/4) — now calls
  `dispatch_render()`, matching `toggle_changed_only`.
- **editor error path corrupts the screen** (`app.rs:151`) + **partial suspend leaves raw
  mode off** (`app.rs:181`) — the Err branch now forces a full clear, and `LiveEditor` always
  attempts to restore the terminal even if suspend or the editor fails.
- **stale state after editor return** (`controller.rs:343`) — I re-render the content pane
  after a successful edit. (Full git status/changed-set refresh-on-return is **deferred**: this
  is a read-only viewer; markers reflect launch/baseline state and the filesystem is the source
  of truth. Tracked as a follow-up, not a v1 blocker.)
- **unconditional redraw / discarded poll effect** (`app.rs:71`, `app.rs:91`) — the loop now
  **draws only when dirty** (driven by `Effects.redraw` and `poll()`'s return), so it no longer
  re-walks the filesystem ~20×/sec while idle.
- **worker panic** (`app.rs:62` info / `controller.rs:143` low) — the worker render is wrapped
  in `catch_unwind`, so a render panic neither stops rendering forever nor fires the **global**
  panic hook from the worker thread (which would reset the main terminal mid-session).
- **tight async-test threshold** (`controller_async.rs:83`) — loosened to a robust non-flaky
  bound (still proves non-blocking).

**Dismissed:** the render.rs doubled-timeout (`render.rs:286`) — unchanged PR2 code, **outside
this PR's diff**; a bounded worst-case on an already-degrading path. Logged as a follow-up.

All fixes are committed; 118 tests pass, PR4 modules clippy-clean, release builds. Round 2 will
verify each finding is resolved at HEAD with no regression.

smarzban:
## Review Gate — Round 2

Findings: 2 total — 0 critical · 0 high · 2 medium · 0 low · 0 info.
_Reviewed by:_ tools + verify-gpt · models: codex:gpt-5.5, deterministic

### Progress since Round 1
✅ Resolved (5): \[correctness\] editor hand-off error path leaves the screen corrupted (no forced clear); \[concurrency\] FIFO render worker lets stale jobs block the current selection; \[correctness\] toggle\_ignore leaves content pane stale when selection shifts; \[caching\] Git state cache is not invalidated after returning from the editor; \[performance\] git diff is read unbounded on the input thread before the size cap — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (2): \[correctness\] Git cache still stale after editor return; \[performance\] Renderer timeout still gets a second full budget

### Must fix
- 🟠 **[MEDIUM]** \[correctness\] Git cache still stale after editor return — `src/controller.rs:381` · 1/1 models _(correctness)_
  The successful editor hand-off path only dispatches a content re-render; it still does not refresh \`git.status()\`, \`git.changed\_set(self.baseline)\`, or the tree's status/filter caches. After editing a file, status markers and changed-only visibility can remain stale even though the pane content updates.
  _Fix:_ On \`Ok(true)\` from \`editor.open\`, refresh status and changed-set when \`is\_git\_repo\`, update the tree with \`set\_status\` and \`set\_changed\_only\`, then dispatch the render and redraw.

- **[MEDIUM]** \[performance\] Renderer timeout still gets a second full budget — `src/render.rs:286`
  _Dismissed:_ Out of this PR's diff — src/render.rs is unchanged PR2 code (git diff origin/main..HEAD does not touch render.rs). The doubled per-render budget is a bounded worst-case on an already-degrading renderer path; it affects only the off-thread worker slot, never input responsiveness. Tracked as a render.rs follow-up; not introduced or worsened by PR4.

smarzban:
## Orchestrator review — Round 2 (single-model verification)

**Decision: 🔴 Request changes** (agrees with the gate's BLOCK).

Round 2 was a single-model verification pass (codex `gpt-5.5`, a reliable JSON parser) over
the round-1 fixes + the deterministic scan (clean). It confirmed **all 5 round-1 blockers
resolved with no regression** — except it correctly caught that one of them, the editor-return
state finding, was only **half** fixed: in round 1 I refreshed the content pane after an edit
but **deferred** re-querying git status/changed-set. So markers and changed-only visibility
stayed stale after editing a file. Good catch — that was a real gap in a git-first-class viewer,
not just a nice-to-have.

**Fixed now:** added `refresh_git_state()` (re-query `git.status()` + `git.changed_set(baseline)`,
update the tree's status/changed-only caches) and call it on a successful editor return as well
as at launch. It's a no-op without a repo, and runs only on deliberate, infrequent actions
(launch / editor return) — never the hot navigation path, so it stays consistent with the
off-thread-diff boundary from round 1. Test-proven (`successful_editor_return_refreshes_git_state`).
119 tests pass; clippy-clean; release builds.

**Dismissed (stands):** the render.rs doubled-timeout (`render.rs:286`) — unchanged PR2 code,
outside this PR's diff; a bounded worst-case affecting only the off-thread worker slot, never
input responsiveness. Tracked as a follow-up.

Round 3 will verify the git-state refresh resolves the finding with no regression.

smarzban:
## Review Gate — Round 3

_Reviewed by:_ tools + verify3-gpt · models: codex:gpt-5.5, deterministic

### Progress since Round 2
✅ Resolved (1): \[correctness\] Git cache still stale after editor return — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:
## Orchestrator review — Round 3 (final)

**Decision: ✅ Approve** (agrees with the gate's PASS).

Round 3 verified the round-2 git-state-refresh fix: the verifier (codex `gpt-5.5`) and the
deterministic scan both came back clean — the editor-return finding is resolved with no
regression. The gate **converged** over three rounds (12 findings → 1 carry-forward → 0).

### What this PR implements (T-18…T-20)
- **Session Controller** — turns each `Intent` into coordinated state changes + `Effects`,
  with Git Service / Content Renderer / Editor Launcher behind injected traits. Git-only
  intents are inert without a repo (AC-26); component failures become non-fatal notices; no
  intent mutates a file (AC-N1/N3). Advances AC-5/6/11/16/20/21/26.
- **Off-thread rendering (AC-23)** — selecting a file dispatches a render job to a worker
  thread (which now also fetches the diff); `handle()` never blocks on a slow renderer; `poll()`
  drains finished renders, applying only the latest selection's (stale dropped by seq).
- **main wiring (AC-17/20)** — `app::run()` reads the herdr context → resolves root/git →
  builds the live components → draws (only when dirty) → input → poll, over a ratatui terminal
  with crossterm raw mode + restore (and ratatui's panic hook). pty smoke test green.

### Fixed across review rounds
**Round 1** (5 blocking + advisories): off-thread the git diff (Arc-shared `Send+Sync`
GitService, worker fetches it — nothing git-related on the input thread); worker collapses its
backlog (no FIFO starvation); `catch_unwind` in the worker (a render panic can't stop rendering
or fire the global panic hook from the worker thread); `toggle_ignore` re-renders content;
editor hand-off forces a clear on the error path and always restores the terminal; the run loop
draws only when dirty (no idle filesystem re-walks); loosened the async test's timing bound.
**Round 2** (1 carry-forward): `refresh_git_state()` re-queries status + changed-set after a
successful editor return (and at launch), so markers/changed-only reflect an in-session edit.

### Deferred / not covered (tracked follow-ups, not v1 blockers)
- `render.rs:286` doubled per-render timeout — unchanged PR2 code, outside this PR's diff;
  bounded worst-case affecting only the off-thread worker slot, never input responsiveness.
- Live git refresh on *external* filesystem changes (outside the editor hand-off) — this is a
  read-only viewer; the filesystem/repo is the source of truth and a re-launch/baseline-toggle
  re-reads it. Out of v1 scope.

**State:** 119 tests pass; PR4 modules clippy-clean; release builds; pty smoke test green.
Merging.

PR #12, opened 2026-06-22
Add zoom toggle (z): hide the tree for a full-screen content pane

smarzban:
## What

Adds a **zoom toggle** (`z`) that hides the file tree so the content pane fills the whole frame — for reading a file full-screen without the tree taking space. Press `z` again to restore the two-column layout.

## Why

The split can only be narrowed to a minimum (`>` clamps the tree at ~20%); there was no way to give the content pane the full width. This is the "maximize content" affordance, staying fully within the read-only viewer model (pure layout state — no file/git mutations, no new rendering path).

## How

- **intent / input** — `Intent::ToggleZoom` added to the closed intent set (`ALL` now 16; the exhaustive no-edit match in `intent.rs` is preserved, so AC-N3 still holds) and mapped to the `z` key. Keyboard-completeness (AC-18) is covered by the existing `every_intent_has_at_least_one_key` test.
- **controller** — a `zoomed` flag, pure layout state (no `dispatch_render`). Entering zoom moves focus to the content pane so `↑`/`↓` scroll the now-full-screen file; leaving zoom returns focus to the tree (back to picking files).
- **presenter** — `ViewState.zoomed`; `columns()` returns content-only when zoomed, overriding both the split *and* the narrow-layout focus rule. The shared `geometry()` therefore reports no tree interior and no divider, so mouse hit-testing routes everything to the content pane (no stale tree/divider regions).
- **docs** — README key table + feature intro, and the manual-verification checklist.

## Tests (TDD)

- `controller.rs::zoom_toggle_hides_tree_and_pins_content_focus` — the toggle flips `zoomed`, pins focus to content on entry and restores it to the tree on exit, and surfaces through `view_state()`.
- `presenter.rs::geometry_is_content_only_when_zoomed` — at a wide width, geometry is content-only (no tree interior, no divider).
- `presenter.rs::zoomed_layout_hides_the_tree_and_fills_with_content` — a draw test: the tree rows are gone and the content fills the frame.

Full suite green; no new clippy warnings. (Repo-wide `cargo fmt` drift is pre-existing and untouched — the code is hand-formatted.)

## Scope / not included

- No persistence — zoom is ephemeral session state (consistent with v1).
- `Tab` while zoomed still flips the focus flag (the hidden tree's cursor); benign, noted for manual verification. Left as-is for YAGNI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ hand-off to an *external* editor — the viewer itself only reads.
   anything else is shown as syntax-highlighted content with line numbers. Cycle the view to
   override.
 - **Navigable content** — scroll the content pane in all four directions, toggle line
-  wrapping, and resize the tree/content split; the layout reflows when the pane is resized.
+  wrapping, resize the tree/content split, or zoom (`z`) to hide the tree and read a file
+  full-screen; the layout reflows when the pane is resized.
 - **Keyboard-first** — every function has a key; no mouse required.
 
 ## Install
@@ -142,6 +143,7 @@ the same locally and remotely; it's only *which* keymap fires them across `--rem
 | `Tab` | Move focus between the tree and content columns |
 | `<` / `>` | Narrow / widen the tree column (move the divider) |
 | `w` | Toggle line wrapping for the content pane |
+| `z` | Zoom — hide the tree so the content pane fills the frame; press again to restore the two-column layout |
 | `r` | Refresh git state — pick up changes made outside the viewer (a merge / pull / commit elsewhere) |
 | `q` / `Esc` | Close the viewer and return to the prior pane |
 
--- a/docs/manual-verification.md
+++ b/docs/manual-verification.md
@@ -60,6 +60,8 @@ This procedure verifies the remaining **live-host** behavior.
      select a changed file to see its diff.
    - Press `Tab` to focus the content pane, then scroll with the arrows (`←`/`→` horizontally,
      `↑`/`↓` vertically), `w` to toggle wrapping, and `<`/`>` to resize the split.
+   - Press `z` to zoom: the tree disappears and the content pane fills the whole frame (focus
+     moves to it, so `↑`/`↓` scroll the file); press `z` again to restore the two columns.
    - Press `e` on a file (with `$EDITOR` set) and confirm your editor opens on that file, then
      exit the editor and confirm the viewer redraws cleanly.
    - Resize the pane (e.g. `herdr pane resize`) and confirm the layout reflows to the new size.
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -722,6 +722,13 @@ impl Controller {
     }
 
     fn toggle_focus(&mut self) -> Effects {
+        // While zoomed the tree is hidden, so there is nothing to switch focus to: keep focus
+        // pinned to the content pane (entering zoom set it there). Without this guard, Tab would
+        // move focus to the invisible tree and route j/k to its cursor — silently re-rendering a
+        // different file behind the full-screen content (review-gate R1, 4-model finding).
+        if self.zoomed {
+            return Effects::noop();
+        }
         self.focus = match self.focus {
             Focus::Tree => Focus::Content,
             Focus::Content => Focus::Tree,
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -151,6 +151,9 @@ pub struct Controller {
     /// User override forcing content wrap on regardless of view mode (the `w` toggle), so long
     /// lines in code/diffs can be wrapped on demand. `false` ⇒ the per-mode default applies.
     wrap_override: bool,
+    /// Hide the tree so the content pane fills the frame (the `z` zoom toggle). Pure layout
+    /// state — the selection and rendered content are unchanged.
+    zoomed: bool,
     tree: TreeModel,
     /// Changed-set vs the active baseline, cached; recomputed on a baseline toggle (AC-16).
     changed: BTreeMap<PathBuf, Status>,
@@ -238,6 +241,7 @@ impl Controller {
             content_height: 0,
             split_pct: SPLIT_DEFAULT,
             wrap_override: false,
+            zoomed: false,
             changed: BTreeMap::new(),
             overrides: HashMap::new(),
             content: Text::raw(""),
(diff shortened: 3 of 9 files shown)


smarzban:
## Review Gate — Round 1

Findings: 2 total — 0 critical · 0 high · 1 medium · 0 low · 1 info.
_Reviewed by:_ holistic-codex + holistic-glm + holistic-kimi + holistic-opus + tools · models: ollama:kimi-k2.7-code:cloud, ollama:glm-5.2:cloud, claude:claude-opus-4-8, codex:gpt-5.5, deterministic

### Must fix
- 🟠 **[MEDIUM]** \[correctness\] Tab while zoomed desyncs focus from the visible layout — `src/controller.rs:724` · 4/4 models _(correctness)_
  Zoom forces focus=Content and hides the tree, but toggle\_focus (Tab) still flips focus unconditionally. After \`z\` then \`Tab\`, focus=Tree while the tree stays hidden and content fills the frame: arrow/j/k keys then drive the invisible tree cursor — navigate() calls dispatch\_render(), so the full-screen content silently switches to a different file with no visible cause — and the content border draws as unfocused. It's recoverable but a confusing state reachable by one documented key, and it reintroduces the focus/layout desync class fixed in 3c4be14.
  _Fix:_ Keep focus consistent with the drawn layout while zoomed — e.g. make toggle\_focus a no-op when self.zoomed (return Effects::noop()), or have it exit zoom (set zoomed=false) so the two columns reappear when focus changes. Add a controller test for ToggleFocus-while-zoomed.

### Advisory (non-blocking)
- ⚪ **[INFO]** \[test-coverage\] No test verifies zoom overrides narrow layout — `tests/presenter\_narrow.rs:33` · 1/4 models _(test-coverage)_
  The Presenter comment and ViewState doc say zoom overrides the narrow-layout focus rule, but the existing zoom tests use a 100-column wide frame; a narrow-frame zoom test would lock in that content fills the frame even when the tree would otherwise be visible.
  _Fix:_ Add a presenter test that sets width \< 80, focus = Tree, zoomed = true and asserts the content pane (not the tree) fills the frame.

smarzban:

Ran the full 4-model holistic panel (kimi · glm · opus · codex) + the deterministic scan. No lenses fired (no subprocess / untrusted-input / spec-AC / concurrency surface — this is a pure layout toggle).

**What the panel found, and what I'm doing:**

- 🟠 **Tab while zoomed desyncs focus (medium, all 4 models) — fixing.** I confirmed it in the code: `toggle_focus` flips focus unconditionally, so `z` then `Tab` lands `focus = Tree` while the tree is hidden — `j`/`k` then drive the invisible cursor and `dispatch_render` silently swaps the full-screen file. Mouse can't reach this (zoomed `geometry` exposes no tree region), so the only entry point is `Tab`. Fix: make `toggle_focus` a no-op while `zoomed`, pinning focus to the content pane for the whole zoom session. Adding a controller test (`Tab` during zoom leaves focus on Content). This closes the focus/layout desync class opus flagged.

- ⚪ **No test for zoom-overrides-narrow (info, kimi) — adding the test.** A presenter test at width < 80 with `focus = Tree, zoomed = true` asserting the content pane fills the frame. Cheap, and it locks in the documented override.

- ⚪ **"Stale scroll bounds on the zoom redraw" (low, codex only) — dismissing as advisory.** Code-checked: `toggle_zoom` triggers a redraw, and `set_content_viewport` re-clamps the scroll on that draw; the only gap is a single, self-correcting frame for wrapped content scrolled near the bottom. This is **identical to the pre-existing behavior of `Event::Resize` and the AC-21 narrow focus-toggle** — both change the content width with the same one-frame clamp lag. It's not zoom-specific; a proper fix is a run-loop "redraw-after-reclamp" change that belongs in its own PR, not bundled into a zoom toggle. Logging as a possible follow-up.

**Decision: 🔴 Request changes** — agrees with the gate's `block`. Implementing the `toggle_focus` guard + the two tests now, then a round-2 single-model verification.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ tools + verify-codex · models: codex:gpt-5.5, deterministic

### Progress since Round 1
✅ Resolved (1): \[correctness\] Tab while zoomed desyncs focus from the visible layout — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

Single-model verification of the round-1 fix (not a fresh sweep). The first verifier (claude-opus-4-8) returned a **non-vote** (unparseable / null output) — that's zero coverage, not a pass, so per the gate's non-vote guard I re-ran with a reliable parser (**codex gpt-5.5**). It came back clean: the round-1 finding is **resolved at HEAD** with **no regression**, and the deterministic scan is clean.

### What this PR implements
A zoom toggle (`z`) that hides the tree so the content pane fills the frame; press again to restore the two-column layout. Pure layout state — no file/git mutation, no new rendering path. `Intent::ToggleZoom` joins the closed intent set (no-edit invariant AC-N3 + keyboard-completeness AC-18 preserved); the presenter draws content-only when zoomed, overriding both the split and the narrow-layout focus rule.

- **R1, 4-model medium — Tab while zoomed desynced focus.** `toggle_focus` flipped focus unconditionally, so `z`→`Tab` landed focus on the hidden tree and routed `j`/`k` to the invisible cursor (silently re-rendering a different file). Fixed by making `toggle_focus` a no-op while zoomed; focus stays pinned to the content pane for the whole zoom session (mouse can't reach the tree while zoomed, so this closes the path entirely). Regression test added.
- **R1 info — coverage gap.** Added a presenter test locking in that zoom overrides the narrow (< 80 col) single-column layout.

### Deferred / not covered
- **"Stale scroll bounds on the zoom redraw" (R1 low, single-model) — advisory, not fixed here.** `set_content_viewport` re-clamps the scroll on the redraw `toggle_zoom` triggers; the only gap is a single self-correcting frame for wrapped content scrolled near the bottom — **identical to the pre-existing `Event::Resize` / AC-21 focus-toggle behavior**, not zoom-specific. A general run-loop "redraw-after-reclamp" fix belongs in its own PR.
- No persistence (zoom is ephemeral session state, consistent with v1).

**Decision: ✅ Approve** — agrees with the gate's `pass`. Merging.

PR #16, opened 2026-06-22
Esc backs out of zoom first; double-tap tolerant of column jitter

smarzban:
## What

Two small UX fixes you asked for:

1. **`Esc`/`q` backs out of zoom first.** When zoomed, the close key un-zooms (tree reappears, focus returns to it) instead of quitting — the instinctive "escape the full-screen view". A second press (now un-zoomed) quits (AC-20). So leaving a zoomed file is two presses: un-zoom, then close.
2. **Double-tap now works like double-click.** `is_double_click` matched the *exact cell*; a touchpad double-tap often lands a column or two apart between taps, so legitimate double-taps were dropped. It now matches the same **row** within the window (ignoring the column) — a tree row is one node end-to-end, so any column on it targets that node. Single-tap (select) is unchanged.

## How

- `controller::close_or_unzoom()` handles `Intent::Close`: zoomed → un-zoom + focus tree (redraw, no quit); else quit.
- `is_double_click` now compares the row (`py == pos.1`) within `DOUBLE_CLICK`, not the full cell. Only the tree-row Up path uses the result, so the relaxation can't cause spurious content/divider activation.
- docs: README Keys (`q`/`Esc`, `z`) + the manual-verification zoom step.

## Tests

- `close_backs_out_of_zoom_first_then_quits` — Close while zoomed un-zooms (no quit); again quits.
- `double_tap_on_the_same_row_activates_even_with_column_jitter` — taps at cols 4 then 9 on the same row expand a folder.
- `is_double_click_requires_the_same_row_within_the_window` — same-row/diff-column ✓, diff-row ✗, too-slow ✗.

196 tests green; no new clippy warnings.

## Review

Small change — per the owner's call, this skips the full review-gate panel and gets a single **opus** review instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,9 @@ the same locally and remotely; it's only *which* keymap fires them across `--rem
 | `Tab` | Move focus between the tree and content columns |
 | `<` / `>` | Narrow / widen the tree column (move the divider) |
 | `w` | Toggle line wrapping for the content pane |
-| `z` | Zoom — hide the tree so the content pane fills the frame; press again to restore the two-column layout |
+| `z` | Zoom — hide the tree so the content pane fills the frame; press again (or `q`/`Esc`) to restore the two-column layout |
 | `r` | Refresh git state — pick up changes made outside the viewer (a merge / pull / commit elsewhere) |
-| `q` / `Esc` | Close the viewer and return to the prior pane |
+| `q` / `Esc` | Back out of zoom if zoomed; otherwise close the viewer and return to the prior pane (AC-20) |
 
 `Tab` to the content pane, then the arrow keys (or `h`/`j`/`k`/`l`) scroll it in all four
 directions; `Tab` back to the tree to move between files. Long lines wrap in prose (markdown /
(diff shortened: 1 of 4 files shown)


smarzban:
## Review — opus (lightweight, per owner's call: small change, single-model review not the full gate)

Deterministic scan: clean. opus reviewed the diff and found **one low** finding — a real one:

- ⚪ **Same-row double-click could spuriously activate across panes (low, correctness).** With the column check dropped, `last_click` was recorded for *every* click, and the tree/content panes share row numbers — so a content-pane click followed by a tree click on the same row within the window would register as a double-click and spuriously activate.

**Fixed (`c5d6a65`):** `last_click` now holds **only tree-row clicks** — the Content / Divider / Outside branches (and the empty-area-below-nodes case) clear it — so a double-click can only ever be two consecutive same-row *tree* clicks, never a cross-pane sequence. The column-agnostic same-row match is therefore safe. Regression test added (`a_content_click_then_a_same_row_tree_click_is_not_a_double_click`).

**Verification:** opus non-voted on the re-check (unparseable output — the documented residual), so per the non-vote guard I re-ran with **codex gpt-5.5** → **0 findings**: the cross-pane activation is resolved and no regression (legitimate tree double-click + single-click select still work).

197 tests green; no new clippy warnings.

**Decision: ✅ Approve** — merging.

PR #19, opened 2026-06-22
refactor: dedup git hardening + backfill security tests (audit)

smarzban:
Audit-driven (Themes 2 & 3). **No behavior change** to observable output; one internal consistency improvement + test coverage.

## Theme 2 — one shared hardened git-command builder
`root::git_output` and `git::git_command` each independently scrubbed the repo-redirecting env (`GIT_DIR`/`GIT_WORK_TREE`/…) and set `core.fsmonitor=false` / `core.hooksPath=/dev/null` — two copies of the untrusted-repo hardening that could **drift** (3 models flagged it). Now `git::git_command` is the single `pub(crate)` builder and the Root Resolver builds its queries through it.
- root's `rev-parse` queries additionally gain `--attr-source=<empty-tree>` — harmless (they read no attributes), and the existing root/baseline/worktree tests confirm resolution is unchanged.
- A unit test asserts the builder applies **every** guard (env scrubbing + fsmonitor/hooks + attr-source + `-C`) — the regression guard against future drift.

## Theme 3 — backfill the hardest / most security-relevant paths
- **git status rename/copy parse** (the trickiest parse): a staged rename reports the new path, and the extra NUL field is consumed so following entries don't desync.
- **AC-27 escape neutralizer**: the previously-untested **OSC**, **C1**, **lone-ESC**, and **ESC+single** branches (only CSI/C0/SGR were covered).
- **render worker `catch_unwind`**: a panicking renderer is contained and the worker stays alive (AC-23 resilience).

## Verification
- `cargo clippy --all-targets -- -D warnings` clean · `cargo fmt --check` clean · all 196 tests green.

Full review-gate (holistic ×4 + lens-security ×2) since it touches untrusted-repo hardening.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -229,9 +229,9 @@ impl Controller {
                     } else {
                         None
                     };
-                // A renderer panic must not kill the worker (rendering would stop forever) nor
-                // fire the global panic hook from this thread (it would reset the main thread's
-                // terminal mid-session). Contain it and surface a placeholder instead.
+                // Contain a renderer panic so the worker survives — otherwise the thread would
+                // die and rendering would stop for the rest of the session. The unwind is caught
+                // here and a placeholder is surfaced in place of the failed render.
                 let result = std::panic::catch_unwind(AssertUnwindSafe(|| {
                     content.render(&job.path, job.mode, raw_diff.as_deref())
                 }))
--- a/src/git.rs
+++ b/src/git.rs
@@ -208,11 +208,15 @@ pub fn diff(
     capture_stdout(cmd)
 }
 
-/// Build a `git -C <repo> <args>` command hardened for read-only use against an
-/// **untrusted** repository: `GIT_OPTIONAL_LOCKS=0` stops status/diff from writing the
-/// index (AC-N2); `core.fsmonitor` / `core.hooksPath` are neutralized so a planted
-/// `.git/config` can't run a program during a query.
-fn git_command(repo_root: &Path, args: &[&str]) -> Command {
+/// Build a `git -C <dir> <args>` command hardened for read-only use against an **untrusted**
+/// repository: `GIT_OPTIONAL_LOCKS=0` stops status/diff from writing the index (AC-N2);
+/// `core.fsmonitor` / `core.hooksPath` are neutralized so a planted `.git/config` can't run a
+/// program during a query; and inherited repo-redirecting env (`GIT_DIR`/`GIT_WORK_TREE`/…) is
+/// dropped so queries resolve against `-C <dir>`, not a repository the viewer was launched
+/// against. **This is the single source of that hardening** — the Root Resolver
+/// ([`crate::root`]) builds its queries through this same function, so the guards cannot drift
+/// between the two.
+pub(crate) fn git_command(repo_root: &Path, args: &[&str]) -> Command {
     let mut cmd = Command::new("git");
     cmd.env("GIT_OPTIONAL_LOCKS", "0")
         // Drop inherited repo-redirecting env so queries resolve against `-C <repo>`, not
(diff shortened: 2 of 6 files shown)


smarzban:
## Review Gate — Round 1

Findings: 4 total — 0 critical · 0 high · 1 medium · 2 low · 1 info.
_Reviewed by:_ holistic + lens-security + tools · models: claude:claude-opus-4-8, ollama:kimi-k2.7-code:cloud, codex:gpt-5.5, deterministic

### Must fix
- 🟠 **[MEDIUM]** \[test-coverage\] panic-containment test does not prove worker survival — `tests/controller\_async.rs:245` · 1/3 models _(test-coverage)_
  The second half of the test breaks as soon as \`ctrl.content()\` still contains "content unavailable", but that placeholder is left over from the first panic. It therefore does not wait for or verify that the worker actually processed the second selection's render job, so a regression that killed the worker after the first panic would pass falsely.
  _Fix:_ Use a renderer that panics on its first call but returns normally thereafter, then verify the second selection renders the expected text rather than a placeholder; this proves the worker loop is still alive and functional after a panic.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[maintainability\] comment incorrectly claims catch\_unwind suppresses the panic hook — `src/controller.rs:232` · 1/3 models _(maintainability)_
  The comment says \`catch\_unwind\` prevents the global panic hook from firing, but Rust always invokes the hook on panic; \`catch\_unwind\` only contains the unwind. The new test comment correctly notes the deliberate panic prints to stderr, creating an inconsistency.
  _Fix:_ Reword the comment to state that the panic is contained so it cannot unwind out of the worker thread, rather than claiming the hook is suppressed.
- ⚪ **[LOW]** \[test-coverage\] Async panic test can pass on stale content — `tests/controller\_async.rs:285` · 1/3 models _(test-coverage)_
  After the first panic, the content pane already contains the placeholder, so the second loop breaks immediately without proving that the second render job was processed or that the worker survived. This makes the regression test weaker than its stated guarantee.
  _Fix:_ Before dispatching the second selection, record a distinguishable signal such as a render attempt counter, clear/wait for a different sequence-visible result, or use a content provider that panics once then returns unique text for the second job and assert that unique text appears.
- ⚪ **[INFO]** \[error-handling\] Root resolution now inherits the git ≥2.40 \`--attr-source\` floor — `src/root.rs:55` · 1/3 models _(error-handling)_
  Routing \`git\_output\` through the shared \`git\_command\` adds \`--attr-source=\<empty-tree\>\` to root-resolution queries, which the old runner did not have. On git \<2.40 (which lacks this option), every root query now exits non-zero, so a real repository is misclassified as not-a-repo and roots at cwd instead of the worktree top-level. Impact is minimal — the Git Service already required \`--attr-source\`, so the plugin already needed modern git, and the result is still a graceful plain-browser degradation per AC-26 (and cwd is usually the repo root anyway) — but it is a real new behavioral delta and there is no documented minimum git version.
  _Fix:_ Either document a minimum git version (≥2.40) in techstack.md, or — if pre-2.40 graceful behavior matters — keep the root resolver's queries free of \`--attr-source\` since they never consult gitattributes (e.g. a builder variant without it). No code change is required to merge.

smarzban:
## Orchestrator review — Round 1 — 🔴 Request changes

Full panel ran: holistic on opus, kimi, and codex (glm non-voted twice — surfaced as lost coverage; 3 effective models, not thin) + lens-security ×2 (both clean) + deterministic scan (clean). I read the code behind every finding.

**Agree & fixing:**
- 🟠 **MEDIUM — panic-containment test (kimi; codex flagged the same at low).** Confirmed real: the test only checked for the placeholder, which was *already on screen from the first panic*, so it never proved the worker **survived**. Fixing by panicking on a specific file (`b.rs`) and then rendering a *different* file (`a.rs`) normally — the distinct post-panic content can only arrive if the worker thread lived.
- ⚪ **LOW — `controller.rs` comment.** `catch_unwind` does not suppress the panic hook; the comment was wrong. Correcting it.
- ⓘ **INFO — git ≥2.40 floor.** Routing root resolution through the shared builder adds `--attr-source`, so root now needs git ≥2.40 (consistent with the rest of the Git Service). Documenting it.

**Noted, deferred:** the defense-in-depth gap behind the comment finding — a renderer panic fires the process-wide hook, which tears down the terminal mid-session — is genuinely low-likelihood (`render()` delegates to subprocesses and never unwraps untrusted input). Logged as a follow-up rather than fixed in this PR.

**Security:** lens-security (2 models) confirmed the hardening dedup weakens no guard — the single `git_command` applies every protection, and the new unit test locks that in.

**Decision: 🔴 Request changes** (agrees with the BLOCK verdict). Fixing now → Round 2.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ tools + verify · models: codex:gpt-5.5, deterministic

### Progress since Round 1
✅ Resolved (1): \[test-coverage\] panic-containment test does not prove worker survival — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:
## Orchestrator review — Round 2 — ✅ Approve

Verification (codex, scoped to Round 1's findings + the fix diff) + deterministic scan: **all four findings resolved, no regression.**

**Fixed (cumulative):**
- panic-containment test now proves worker **survival** via distinct post-panic content, not just the placeholder;
- corrected the `catch_unwind` comment (dropped the false hook claim);
- documented the git ≥2.40 floor on root resolution.

**What this PR implements:** one shared hardened git-command builder — eliminating the duplicated untrusted-repo hardening between the Root Resolver and the Git Service, with a unit test that locks in every guard — plus backfilled tests for the hardest paths (git rename/copy parse, the AC-27 OSC/C1/lone-ESC/ESC+single branches, and render-worker panic containment).

**Deferred:** worker-panic terminal safety (low-likelihood defense-in-depth, logged); the broader test-hygiene / UX / a11y items from the audit backlog.

**Verification:** `clippy -D warnings` clean · `fmt --check` clean · all tests green.

**Decision: ✅ Approve** (agrees with the PASS verdict).

PR #28, opened 2026-06-22
fix(viewer): root the tree at the invoking pane's dir, not the plugin's

smarzban:
**Bug (hit on first real use):** the file tree showed the viewer's *own* install directory instead of the directory the user invoked it from.

**Cause:** herdr starts the pane from the plugin root (the `[[panes]]` command is a relative path `./target/release/…`), so the viewer's process cwd is the plugin dir. It fell back to that because it looked for a `cwd` field in `HERDR_PLUGIN_CONTEXT_JSON` — but **herdr 0.7.0 actually provides the invoking pane's directory as `focused_pane_cwd`** (and the workspace root as `workspace_cwd`). Confirmed by dumping the real injected context.

**Fix:** read `focused_pane_cwd` → `workspace_cwd` → `cwd` → process cwd (most specific wins). One-field change in the host-context parser; no launcher change needed (herdr already injects the right info).

**Verified live in herdr:** opened the viewer next to a pane in a marker dir → the tree shows that dir's files (`PROJECT_MARKER.txt`, `notes.md`, `src/`), not the plugin's. Plus unit tests for the real 0.7.0 context shape. fmt/clippy clean, all tests green.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -36,10 +36,13 @@ pub fn parse_context(json: Option<&str>, fallback_cwd: PathBuf) -> LaunchContext
     let raw: RawContext = json
         .and_then(|s| serde_json::from_str(s).ok())
         .unwrap_or_default();
+    // Ignore empty-string fields (a malformed host value) so they fall through to the next
+    // candidate / the process-cwd fallback rather than rooting at an empty path.
     let cwd = raw
         .focused_pane_cwd
-        .or(raw.workspace_cwd)
-        .or(raw.cwd)
+        .filter(|s| !s.is_empty())
+        .or(raw.workspace_cwd.filter(|s| !s.is_empty()))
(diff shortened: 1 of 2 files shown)


smarzban:
✅ **Opus review** — APPROVE. Correct, backward-compatible (legacy `cwd` still works), and **no trust-boundary widening** (the value comes from herdr's injected context, the trusted host — not from viewed file content; it flows into the existing read-only `git rev-parse` + within-root path checks). AC-26 'malformed/absent → fallback, never panic' holds. **Verified live in herdr** + 9 unit tests.

Addressed both non-blocking suggestions: ignore empty-string context fields (degrade to the fallback rather than root at an empty path) and added an explicit precedence test (`focused_pane_cwd` beats a co-present legacy `cwd`).

PR #30, opened 2026-06-22
feat: update-available notification (v1.1.0)

smarzban:
## What & why

herdr has no plugin auto-update and no new-release notification, so users have no way to learn that a newer release exists. This adds an unobtrusive, opt-out **update-available banner**: when the viewer launches and a newer release exists, a one-row bottom status line names the new version and the one-line update command.

## How

- **`src/update/`** (new module, no new dependencies):
  - `version.rs` — hand-rolled `Version` + `git ls-remote --tags` parsing (skips `^{}` peels and pre-releases), `current`/`latest_stable`/`newer_than_current`.
  - `cache.rs` — throttle cache (`should_check` 24h window, `next_cache`, XDG cache dir, best-effort load/store). Stores only a unix timestamp + version string.
  - `mod.rs` — `git ls-remote` probe (injected runner; git low-speed abort so it can't hang), banner text, `decide` (pure startup decision), and a background-thread `start_with`/`start_default` that mirrors the existing off-thread render pattern (`std::thread` + `mpsc`, drained by `Controller::poll`).
- **Controller** — banner state set from the cached value immediately, refreshed by the background result; `u` dismisses for the session (not persisted; returns next launch while still behind).
- **Presenter** — a one-row bottom status line, drawn only when behind; `draw`/`geometry` carve the same body so mouse hit-testing stays correct.
- **Opt-out** — `HERDR_FILE_VIEWER_NO_UPDATE_CHECK=1` disables the check and banner entirely; the e2e harness sets it so tests never hit the network.

## Properties

Read-only (only network action is `git ls-remote` against our own repo), non-blocking (off the UI thread), throttled (≤ once / 24h), fail-silent (offline / git-missing / no-tags → no banner, no error), zero new dependencies.

## Docs

README gains an **Updating** section and de-emphasizes `--ref` (a bare `herdr plugin install …` already pulls the latest; `--ref` only pins an older version). Bumps to **v1.1.0**.

All tests + clippy `-D warnings` + fmt green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to this project are documented here. The format is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.0] - 2026-06-22
+
+### Added
+- **Update-available notification** — the viewer checks for a newer release (at most once per
+  day, off the UI thread, over a read-only `git ls-remote`) and, when you're behind, shows a
+  dismissable bottom status-line banner with the one-line update command. Press `u` to dismiss
+  it for the session. Opt out entirely with `HERDR_FILE_VIEWER_NO_UPDATE_CHECK=1`. No new
+  dependencies, no telemetry.
+
+### Docs
+- Clarified updating: re-running `herdr plugin install …` pulls the latest; `--ref` only pins a
+  specific version and is no longer presented as part of the normal install.
+
 ## [1.0.0] - 2026-06-22
 
 First public release: a git-aware, read-only file viewer that runs as a herdr plugin pane.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -574,7 +574,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "herdr-file-viewer"
-version = "1.0.0"
+version = "1.1.0"
 dependencies = [
  "ansi-to-tui",
  "crossterm",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "herdr-file-viewer"
-version = "1.0.0"
+version = "1.1.0"
 edition = "2024"
 rust-version = "1.96"
 description = "A git-aware, read-only file viewer that runs as a herdr TUI pane"
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ the launcher's open/focus/toggle behavior, the full key map, and the `--remote` 
 | `w` | Toggle line wrapping for the content pane |
 | `z` | Zoom — hide the tree so the content pane fills the frame; press again (or `q`/`Esc`) to restore the two-column layout |
 | `r` | Refresh git state — pick up changes made outside the viewer (a merge / pull / commit elsewhere) |
+| `u` | Dismiss the "update available" banner for this session |
 | `q` / `Esc` | Back out of zoom if zoomed; otherwise close the viewer and return to the prior pane |
 
 `Tab` to the content pane, then the arrow keys (or `h`/`j`/`k`/`l`) scroll it in all four
@@ -159,9 +160,9 @@ Requirements: **Rust 1.96+** (edition 2024) and Cargo; **herdr 0.7.0+**, on **Li
 which the viewer pane launches:
 
 ```bash
-# install from this GitHub repo (latest main):
+# install (and update — re-run any time to get the latest):
 herdr plugin install smarzban/herdr-file-viewer
-# …or pin a specific release for reproducibility:
+# …optional: pin a specific older version for reproducibility:
 herdr plugin install smarzban/herdr-file-viewer --ref v1.0.0
 
 # or, for local development, link this checkout in place:
@@ -169,11 +170,34 @@ cargo build --release            # plugin link does NOT run the [[build]] step, 
 herdr plugin link /path/to/herdr-file-viewer
 ```
 
+> You don't need `--ref` to stay current — a bare install pulls the latest. See
+> [Updating](#updating).
+
 Confirm it registered with `herdr plugin list`. To build manually outside herdr:
 
 ```bash
 cargo build --release
 ```
+
+## Updating
+
+herdr has no plugin auto-update, so the viewer tells you when a new release exists: open it
+(`prefix+f`) and, if you're behind, a status line appears at the bottom naming the new version
+and the command to update. Press `u` to dismiss it for the session.
+
+To update, just re-run the install — it pulls the latest:
+
+```bash
+herdr plugin install smarzban/herdr-file-viewer
+```
+
+- You **don't** need `--ref` to stay current; it only *pins* a specific version (and a pin stays
+  pinned until you change it).
+- Want a heads-up the moment a release ships? On GitHub, **Watch → Custom → Releases**.
+- Prefer no network check? Set `HERDR_FILE_VIEWER_NO_UPDATE_CHECK=1` in the pane's environment —
+  the check (and banner) are disabled entirely. The check otherwise runs at most once per 24h,
+  off the UI thread, over a read-only `git ls-remote`, and never blocks or fails the viewer when
+  offline.
 
 ## Optional runtime dependencies (external renderers)
 
--- a/herdr-plugin.toml
+++ b/herdr-plugin.toml
@@ -6,7 +6,7 @@
 
 id = "herdr-file-viewer"
 name = "herdr-file-viewer"
-version = "1.0.0"
+version = "1.1.0"
 description = "A git-aware, read-only file viewer: a keyboard-driven TUI in a herdr split pane."
 min_herdr_version = "0.7.0"
 platforms = ["linux", "macos"]
--- a/src/app.rs
+++ b/src/app.rs
@@ -72,6 +72,9 @@ pub fn run() -> io::Result<()> {
             editor,
         },
     );
+    // Kick off the once-a-day update check (off the UI thread; disabled by
+    // HERDR_FILE_VIEWER_NO_UPDATE_CHECK). The banner, if any, appears on a later draw.
+    controller.set_update(crate::update::start_default());
 
     let mut terminal = ratatui::try_init()?;
     // Mouse is additive to the keyboard-first design (AC-18): herdr forwards mouse events to a
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -20,6 +20,7 @@ use crate::git::{Baseline, Status};
 use crate::intent::Intent;
 use crate::presenter::{Focus, PaneGeometry, ViewState};
 use crate::tree::{Node, NodeKind, TreeModel};
+use crate::update::{self, UpdateState, Version};
 use crate::view_policy::{FileDescriptor, ViewMode, applicable_modes, default_mode};
 use crossterm::event::{KeyModifiers, MouseButton, MouseEvent, MouseEventKind};
 use ratatui::layout::Position;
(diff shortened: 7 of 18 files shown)


smarzban:
## Review Gate — Round 1

Findings: 8 total — 0 critical · 2 high · 1 medium · 5 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-simplify + lens-spec + lens-subtle-correctness + tools · models: claude:claude-opus-4-8, codex:gpt-5.5, ollama:glm-5.2:cloud, ollama:kimi-k2.7-code:cloud, deterministic

### Must fix
- 🔴 **[HIGH]** \[security\] update probe reads untrusted repo git config — `src/update/mod.rs:62` · 2/4 models _(security)_
  The new startup \`git ls-remote\` inherits the viewer's current working directory and does not scrub repo-local Git config, so opening a malicious repository can influence this network probe via settings like URL rewrites and protocol/SSH helpers. This breaks the existing hardening model for Git invocations and can turn a read-only update check into attacker-controlled command execution or an unintended outbound request.
  _Fix:_ Build the update probe with the same security boundary as other Git calls: run it from an app-controlled non-repository directory, remove repo-redirecting Git env vars, and force safe config such as disabling \`protocol.ext\`/unexpected protocols before \`ls-remote\`.
- 🔴 **[HIGH]** \[security\] Update check runs \`git ls-remote\` in the untrusted viewed repo's cwd, so a malicious \`.git/config\` redirects it — `src/update/mod.rs:62` · 2/4 models _(security)_
  \`run\_git\_ls\_remote\` spawns \`git\` inheriting the process cwd, which is the viewed directory (an untrusted git repo in the tool's primary use case, per the markdown-style threat model). git loads the local repo's \`.git/config\` for \`git ls-remote \<explicit-url\>\`, so a one-line \`url.\*.insteadOf\` (or \`http.proxy\`/credential-helper) entry redirects the once-a-day update check to an attacker server — leaking the user's IP/timing and letting the attacker forge the \`git ls-remote\` response (an arbitrary banner version). I verified this: a malicious \`.git/config\` with \`insteadOf\` made \`git ls-remote https://github.com/smarzban/herdr-file-viewer\` resolve \`attacker.example.invalid\` instead.
  _Fix:_ Run the probe from a neutral non-repo directory so git never discovers the untrusted local config, e.g. \`.current\_dir(cache\_dir().unwrap\_or\_else(std::env::temp\_dir))\` and \`.env("GIT\_CONFIG\_NOSYSTEM", "1")\` (and unset \`GIT\_CONFIG\_GLOBAL\` to \`/dev/null\` if you want to fully isolate). The compiled-in URL is already trusted; only the cwd-borne local config is the hole.
- 🟠 **[MEDIUM]** \[correctness\] Failed probes still rewrite the throttle cache — `src/update/mod.rs:158` · 2/4 models _(correctness)_
  The spec's error table says offline, git-missing, non-zero, or hanging probes leave the cache untouched, but this stores \`next\_cache(...)\` even when \`succeeded\` is false. On a first offline launch this writes a fresh \`last\_check\_unix\` with no version, suppressing another update check for 24h after connectivity returns.
  _Fix:_ Only call \`cache::store\` after a successful probe result; on probe failure, send nothing and leave the existing cache file unchanged.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[error-handling\] update\_rx is never cleared on the probe-failure path, so poll() polls a disconnected channel every tick for the session — `src/controller.rs:997` · 2/4 models _(error-handling)_
  The background thread only sends on the channel when the probe succeeds (start\_with: \`if succeeded { tx.send(...) }\`). On a probe failure it sends nothing and drops the sender, so the channel disconnects with no message. poll() then gets Err(Disconnected) from try\_recv(), never enters the block, and leaves update\_rx as Some forever — so every event-loop tick (\~20×/s) calls try\_recv() on a dead channel for the rest of the session. Harmless but it's the one teardown path the new receiver state isn't cleaned up on.
  _Fix:_ Match on try\_recv() and clear the receiver on both Ok and Err(Disconnected): e.g. \`match rx.try\_recv() { Ok(v) =\> { self.update\_available = v; self.update\_rx = None; applied = true; } Err(TryRecvError::Disconnected) =\> self.update\_rx = None, Err(TryRecvError::Empty) =\> {} }\`.
- ⚪ **[LOW]** \[correctness\] Herdr-provided plugin cache dir is ignored — `src/update/cache.rs:53` · 1/4 models _(correctness)_
  The cache resolution requirement says to prefer a herdr-provided plugin dir when exposed, then fall back to XDG/HOME. \`cache\_dir()\` only checks \`XDG\_CACHE\_HOME\` and \`HOME\`, so it can store the throttle cache in the wrong location under a host that provides a plugin directory.
  _Fix:_ Read the herdr-exposed plugin/cache directory from the runtime context or environment first, and fall back to the current XDG/HOME behavior only when absent.
- ⚪ **[LOW]** \[caching\] Lost-update race on the shared throttle cache across concurrent instances — `src/update/cache.rs:69` · 1/4 models _(caching)_
  The cache is one global file shared by every viewer process (herdr is multi-pane). Each instance does an unsynchronized read-modify-write: it loads \`latest\_seen\` at startup, and \`next\_cache\` re-emits that loaded value on a probe that fails or finds nothing (cache.rs:39-43), then \`store\` blindly truncates+overwrites (non-atomic \`fs::write\`). Interleaving: instance A probes successfully and writes latest\_seen=2.0.0; instance B (loaded the older value, then its probe fails/finds-nothing) writes last and reverts latest\_seen to the stale value while advancing the timestamp. Because the timestamp advanced, \`should\_check\` now suppresses re-checks for 24h, so the just-discovered banner is silently lost/stale for a full window. Self-corrects next window; impact is an advisory banner, hence low.
  _Fix:_ Before storing, re-read the on-disk cache and keep the newer of (disk latest\_seen, probed latest) rather than the value loaded at startup — i.e. only ever advance latest\_seen. Pair with an atomic write (write to a temp file in the same dir, then \`fs::rename\`) so a concurrent writer can never see a torn file either.
- ⚪ **[LOW]** \[error-handling\] git ls-remote connect/DNS phase is not bounded by the low-speed settings the comment relies on — `src/update/mod.rs:28` · 1/4 models _(error-handling)_
  The comment claims the low-speed env vars stop a \`git\` child from "hanging indefinitely," but GIT\_HTTP\_LOW\_SPEED\_LIMIT/TIME only abort stalled HTTP transfers — they do not bound TCP connect or DNS resolution. On a black-holed network the connect can hang for the OS default (tens of seconds to minutes). It runs on a detached thread so it never blocks the UI, but after the viewer quits the orphaned git child can linger past the implied 5s bound.
  _Fix:_ Either soften the comment to scope the guarantee to transfer stalls, or harden it (e.g. wrap with a timeout / a hard kill of the child after a deadline) if a strictly-bounded child is desired.
- ⚪ **[LOW]** \[test-coverage\] Fresh-cache no-network path lacks a spy test — `tests/update\_banner.rs:76` · 1/4 models _(test-coverage)_
  AC-U4 requires a fresh cache to show the cached banner with no network call, but the tests only cover the pure \`decide\` result and do not verify the startup/session path leaves the probe runner uncalled. A regression that spawned a probe despite a fresh cache could pass the current session tests.
  _Fix:_ Add a \`start\_with\` or controller startup test using a fresh cached newer version and a spy or panic probe, asserting the banner is present, \`rx\` is \`None\`, and the probe call count stays zero.

smarzban:
## Orchestrator review — Round 1 🔴 Request changes

I ran the full discovery panel: **holistic** on opus-4-8, gpt-5.5, glm-5.2 (kimi-k2.7's holistic pass came back empty twice — a flaky non-result; **that one holistic vote is lost coverage**, though kimi did vote on `lens-subtle-correctness`), plus fired lenses **security**, **subtle-correctness**, **spec**, **simplify** across diverse models, and the deterministic scan (clean; gitleaks/osv-scanner not installed → secret/dep scanning skipped, but the diff adds no secrets and no new deps).

**Agreeing and fixing:**

- 🔴 **HIGH — untrusted-repo git config (3 passes / 2 lineages, glm live-verified it).** `git ls-remote` inherits the process cwd and reads repo-local `.git/config`; a malicious `url.*.insteadOf` (or credential-helper) can redirect the probe → IP/timing leak, forged banner version, or worse. This breaks the project's own untrusted-repo threat model (the same one `markdown_style` already honors). **Fix:** run the probe from a neutral non-repo directory with repo discovery disabled (`GIT_CEILING_DIRECTORIES`), restrict to `https`, keep terminal-prompt off — so no repo-local config is ever read regardless of launch cwd.
- 🟠 **MEDIUM — failed probes rewrite the throttle cache (2 models).** Real, and it contradicts the spec's own error table. **Fix:** persist the cache only on a *successful* probe; a failure leaves it untouched so the check retries next launch.
- ⚪ Also fixing these worthwhile lows: clear `update_rx` on channel-disconnect; bound the git child with a hard wall-clock timeout (connect/DNS isn't covered by the low-speed settings) + scope the comment honestly; atomic cache write (`tmp`+`rename`) for concurrent panes; inline the thin `probe` wrapper; add the AC-U4 spy test (fresh cache ⇒ probe never called).

**Dismissed (with code-checked reason):**

- ⚪ *"Herdr-provided plugin cache dir is ignored."* Verified: herdr 0.7.0's injected context exposes `focused_pane_cwd`/`workspace_cwd`/`tab_id` only — no cache/config dir — and `herdr plugin config-dir` is CLI-only, not surfaced to the running process. There is no runtime hook to read. XDG cache is the correct, intentional home for a throttle cache. Not a bug.

**Decision: 🔴 Request changes.** Implementing the above now; Round 2 will verify each at HEAD.

smarzban:
## Review Gate — Round 4

_Reviewed by:_ tools + verify · models: ollama:glm-5.2:cloud, deterministic

### Progress since Round 3
✅ Resolved (3): \[security\] update probe reads untrusted repo git config; \[security\] Update check runs \`git ls-remote\` in the untrusted viewed repo's cwd, so a malicious \`.git/config\` redirects it; \[correctness\] Failed probes still rewrite the throttle cache — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:
## Orchestrator review — Final (Round 4) ✅ Approve

Converged to **PASS** after a round-1 discovery panel and three verification rounds. The security
boundary around the new `git ls-remote` probe was hardened iteratively until two independent
lineages (codex → glm) confirmed it closed.

**What this PR implements:** an opt-out, once-a-day, off-the-UI-thread update-available check that
shows a dismissable bottom status-line banner when a newer release exists, with the one-line update
command. No new dependencies; read-only; fail-silent.

**Round-by-round:**

- **R1 — discovery (holistic ×3 lineages + lens-security/spec/subtle-correctness/simplify + scan).**
  Found a **HIGH**: the probe inherited the process cwd and read repo-local `.git/config`, so a
  malicious viewed-repo `url.insteadOf`/credential-helper could redirect or hijack it (glm
  live-verified the exploit). Plus a **MEDIUM** (failed probes rewrote the throttle cache, against
  the spec) and lows. *(One holistic vote — kimi — was lost to a flaky empty result; covered by 3
  other holistic lineages + diverse lenses.)*
- **R2 — verify (codex).** Caught a residual in R1's fix: running in the shared system temp dir
  still let git read a `.git/config` in that cwd. → fixed by running from a fresh private dir.
- **R3 — verify (codex).** Caught a further residual: the dir used a predictable path with
  non-exclusive creation, so a planted/raced dir could be reused. → fixed with exclusive creation
  (`std::fs::create_dir`) + an unguessable, never-reused name.
- **R4 — verify (glm-5.2, independent lineage).** **Clean** — residual resolved, no regression.

**Fixed (cumulative):**
- 🔴 HIGH probe-config hijack — now runs `git ls-remote` from a freshly-, exclusively-created,
  unguessably-named **private empty dir** with discovery ceiling-bounded and transport pinned to
  `https`; never prompts. Hermetic `--get-url` regression test proves a malicious repo-local
  `insteadOf` is ignored.
- 🟠 MEDIUM cache — throttle cache is written only on a **successful** probe; failures retry next
  launch.
- ⚪ lows — `update_rx` cleared on channel-disconnect; whole probe bounded by a hard wall-clock
  timeout (connect/DNS, not just transfer); atomic cache write for concurrent panes; thin `probe`
  wrapper inlined; AC-U4 fresh-cache spy test added.

**Dismissed (code-checked):** "herdr-provided cache dir ignored" — herdr 0.7.0 exposes no such
runtime env var; XDG cache is the correct, intentional home.

**Deferred / not covered:** none blocking. The probe keeps the user's own global/system git config
(needed for proxy/CA setups; only the *viewed repo* is the untrusted surface). Prebuilt-binary
release pipeline remains separately parked.

All 24 test binaries, `clippy -D warnings`, and `fmt` green at HEAD.

**Decision: ✅ Approve** — merging.

PR #31, opened 2026-06-23
feat: prebuilt-binary install path (no Rust toolchain needed) — v1.2.0

smarzban:
## What & why

Lets users install the plugin **without a Rust toolchain**. At install time herdr's `[[build]]` step now runs `scripts/fetch-or-build.sh`, which downloads a **SHA-256-verified prebuilt binary matching the source's version + platform** and falls back to `cargo build --release` on any miss. The install command is unchanged and never harder than before. Ships as **v1.2.0**.

## How it works

- **`scripts/fetch-or-build.sh`** (new `[[build]]` command): resolves the target triple from `uname`, reads the version from `Cargo.toml`, downloads `…/releases/download/v<version>/herdr-file-viewer-<triple>` + `SHA256SUMS` over HTTPS, **verifies the hash before `chmod +x`**, and installs to `target/release/herdr-file-viewer`. **Any** failure (unsupported platform, missing asset, network error, checksum mismatch, no curl/wget) prints a stderr notice and execs the original cargo build (preserving the `~/.cargo/env` sourcing + "install Rust 1.96+" message).
- **`.github/workflows/release.yml`** (new): on a `v*` tag push, cross-builds the three targets — `aarch64-apple-darwin` + `x86_64-apple-darwin` (one macOS runner) and `x86_64-unknown-linux-musl` (static, ubuntu) — guards that the tag matches `Cargo.toml`'s version, and publishes the binaries + `SHA256SUMS` to the release (create-or-update + `--clobber`).
- **Version-matching (load-bearing):** the script fetches the tag `v<Cargo.toml version>` **exactly, never "latest"** — so the binary always matches the cloned source. A tagged commit hits the fast path; a between-release `main` HEAD cleanly falls back to a source build.

## Tests

- `tests/fetch_or_build.rs` — 7 hermetic tests (PATH-stubbed `uname`/`curl`/`cargo`, real `sha256sum`/`shasum`, local fixture): per-platform triple mapping, checksum-verified install, and mismatch / missing-asset / unmapped-platform → fallback (asserts nothing unverified is ever installed).
- `tests/release_workflow.rs` — 4 string-contract tests (mirrors `tests/manifest.rs`): tag trigger, the three targets, `SHA256SUMS`, and the tag-vs-`Cargo.toml` guard. The YAML was also validated with a real `yaml.safe_load`.
- Full suite: 239 passing; clippy `-D warnings` and `cargo fmt --check` clean.

## Scope / notes

- **No new Cargo dependencies.** Read-only plugin — no runtime behavior change; crosses no constitution line (no ADR).
- **Out of scope (deferred):** signatures (layer on later with zero install-UX change), Windows, ARM-Linux, tar.gz packaging.
- **Follow-up logged (AUDIT.md):** GitHub Actions are pinned by version tag, not commit SHA — consistent with the existing `ci.yml`; a repo-wide SHA-pin is a deferred hardening pass (owner decision).

## Verification after merge

Cut `v1.2.0` (owner-gated) → the new workflow attaches the 3 binaries + `SHA256SUMS`; then confirm a Rust-less `herdr plugin install` reports `installed prebuilt v1.2.0 …`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ on:
     tags: ["v*"]
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   build:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,94 @@
+name: Release
+
+on:
+  push:
+    tags: ["v*"]
+
+permissions:
+  contents: write
+
+jobs:
+  build:
+    name: build ${{ matrix.triple }}
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - os: macos-latest
+            triple: aarch64-apple-darwin
+          - os: macos-latest
+            triple: x86_64-apple-darwin
+          - os: ubuntu-latest
+            triple: x86_64-unknown-linux-musl
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+
+      # Fail fast if the pushed tag disagrees with the crate version, so we never publish a release
+      # whose assets don't match the source the install step clones.
+      - name: Verify tag matches Cargo.toml version
+        shell: bash
+        run: |
+          tag="${GITHUB_REF_NAME#v}"
+          crate="$(grep -E '^version *= *"' Cargo.toml | head -n1 | sed -E 's/^version *= *"([^"]+)".*/\1/')"
+          if [ "$tag" != "$crate" ]; then
+            echo "tag $GITHUB_REF_NAME ($tag) != Cargo.toml version ($crate)" >&2
+            exit 1
+          fi
+
+      - uses: dtolnay/rust-toolchain@1.96.0
+        with:
+          targets: ${{ matrix.triple }}
+
+      - name: Install musl tools
+        if: matrix.triple == 'x86_64-unknown-linux-musl'
+        run: sudo apt-get update && sudo apt-get install -y musl-tools
+
+      - name: Build
+        run: cargo build --release --target ${{ matrix.triple }}
+
+      - name: Stage asset + checksum
+        shell: bash
+        run: |
+          mkdir -p dist
+          asset="herdr-file-viewer-${{ matrix.triple }}"
+          cp "target/${{ matrix.triple }}/release/herdr-file-viewer" "dist/$asset"
+          cd dist
+          if command -v sha256sum >/dev/null 2>&1; then
+            sha256sum "$asset" > "$asset.sha256"
+          else
+            shasum -a 256 "$asset" > "$asset.sha256"
+          fi
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.triple }}
+          path: dist/*
+
+  publish:
+    name: publish release
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          path: artifacts
+
+      - name: Assemble release assets + SHA256SUMS
+        shell: bash
+        run: |
+          mkdir -p release
+          # Flatten the per-target binaries and concatenate their checksum lines into one file.
+          find artifacts -type f -name 'herdr-file-viewer-*' ! -name '*.sha256' -exec cp {} release/ \;
+          cat artifacts/*/*.sha256 > release/SHA256SUMS
+          echo "Publishing:"; ls -l release; echo "---"; cat release/SHA256SUMS
+
+      - name: Create-or-update the release and upload assets
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release view "$GITHUB_REF_NAME" --repo "$GITHUB_REPOSITORY" >/dev/null 2>&1 \
+            || gh release create "$GITHUB_REF_NAME" --repo "$GITHUB_REPOSITORY" --title "$GITHUB_REF_NAME" --generate-notes
+          gh release upload "$GITHUB_REF_NAME" release/* --clobber --repo "$GITHUB_REPOSITORY"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project are documented here. The format is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.0] - 2026-06-23
+
+### Added
+- Prebuilt-binary install path: tagged releases now ship SHA-256-verified binaries for macOS
+  (arm64 + x86_64) and Linux x86_64 (static/musl). The install step downloads the binary matching
+  the source's version + platform and falls back to a `cargo` source build on any miss, so no Rust
+  toolchain is needed for supported platforms. The install command is unchanged.
+
 ## [1.1.0] - 2026-06-22
 
 ### Added
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -574,7 +574,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "herdr-file-viewer"
-version = "1.1.0"
+version = "1.2.0"
 dependencies = [
  "ansi-to-tui",
  "crossterm",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "herdr-file-viewer"
-version = "1.1.0"
+version = "1.2.0"
 edition = "2024"
 rust-version = "1.96"
 description = "A git-aware, read-only file viewer that runs as a herdr TUI pane"
--- a/README.md
+++ b/README.md
@@ -54,13 +54,8 @@ right into the tree. It opens beside whatever you're doing and never touches you
 
 ## Quick start
 
-> **Prerequisite: [Rust](https://rustup.rs) 1.96+.** The plugin compiles from source at install
-> time, so `cargo` must be on your PATH. No Rust yet?
-> `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh` (or `brew install rust`).
-> If it's missing, the install fails with a message telling you exactly this.
-
 ```bash
-# 1. Install the plugin (herdr builds it from source at install time):
+# 1. Install the plugin (tagged releases download a prebuilt binary; otherwise builds from source):
 herdr plugin install smarzban/herdr-file-viewer
 
 # 2. (recommended) install the renderers, so markdown / diffs / code are styled, not plain text:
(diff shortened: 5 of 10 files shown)


smarzban:
## Review Gate — Round 1

Findings: 9 total — 0 critical · 0 high · 4 medium · 5 low · 0 info.
_Reviewed by:_ holistic + lens-security + tools · models: deterministic, codex:gpt-5.5, ollama:glm-5.2:cloud, ollama:kimi-k2.7-code:cloud, claude:claude-opus-4-8

### Must fix
- 🟠 **[MEDIUM]** \[security\] build job inherits unnecessary contents:write permission — `.github/workflows/release.yml:7` · 2/4 models _(security)_
  Top-level \`permissions: contents: write\` applies to the build matrix jobs, which only check out code, build, and upload artifacts; they never need to mutate releases or repository contents. Granting write access to jobs that compile arbitrary dependency code widens the blast radius of a supply-chain or build-script compromise.
  _Fix:_ Set top-level permissions to \`contents: read\` (enough for checkout), and keep the explicit \`contents: write\` only on the \`publish\` job that actually creates the release and uploads assets.
- 🟠 **[MEDIUM]** \[correctness\] Version-keyed prebuilt serves a stale release binary when main is ahead, contradicting the documented fallback — `scripts/fetch-or-build.sh:67` · 3/4 models _(correctness)_
  The script downloads the prebuilt for whatever version Cargo.toml declares (\`v$version\`). This repo's convention is that main sits at the last released version between releases (main is currently 1.1.0 with a v1.1.0 tag, no dev suffix). A bare \`herdr plugin install smarzban/herdr-file-viewer\` clones main HEAD; once main advances past a release with the same Cargo.toml version, the script finds the matching \`v$version\` release and silently installs the previous release's binary instead of main's HEAD source — a behavior regression from the old always-build-from-source step, where main installs got main's code. README.md:155 explicitly promises 'installing from a main that is ahead of the latest release — it automatically builds from source', which is false in exactly this (normal, recurring) case.
  _Fix:_ Make source==binary an invariant of the fast path: either bump main's Cargo.toml to a next-dev pre-release version immediately after each release (e.g. \`1.3.0-dev\`) so non-release sources never match a published asset and genuinely fall back to source as documented; or gate the prebuilt fetch on the checked-out commit actually being the \`v$version\` tag (compare git rev when available). Otherwise correct README.md to state the prebuilt tracks the version number and may lag an unreleased main.
- 🟠 **[MEDIUM]** \[correctness\] Prebuilt fetch can install a stale binary for newer source — `scripts/fetch-or-build.sh:67` · 1/4 models _(correctness)_
  The fast path selects the release asset solely from Cargo.toml's version, so a default-branch or --ref install whose source has moved past tag v$version but has not bumped Cargo.toml will silently install the old tagged binary instead of the checked-out source. That can break later manifest/script changes that expect the current binary, and it contradicts the documented fallback for main ahead of the latest release.
  _Fix:_ Only use the prebuilt asset when the checkout is exactly the matching v$version tag (for example via git rev-parse/describe), otherwise fall back to cargo build; add a test for same-version source ahead of the tag.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[security\] Install script trusts environment overrides in production — `scripts/fetch-or-build.sh:21` · 1/4 models _(security)_
  FV\_BASE\_URL, FV\_OUT, and FV\_CARGO\_TOML are read from the environment without validation. An attacker who can influence the herdr build environment can redirect the download to an attacker-controlled server, alter the resolved version, or write the installed binary to an arbitrary path.
  _Fix:_ Gate these overrides behind an explicit test/debug flag (e.g., require FV\_TEST\_MODE=1) so they are not honored during a normal user install.
- ⚪ **[LOW]** \[test-coverage\] wget and shasum fallback paths are not exercised — `scripts/fetch-or-build.sh:70` · 1/4 models _(test-coverage)_
  The script has fallback logic for wget and shasum, but the hermetic test always stubs curl and uses sha256sum. A regression in the wget or shasum branches would not be caught by CI.
  _Fix:_ Add two focused tests that remove curl from PATH (leaving a wget stub) and remove sha256sum from PATH (leaving a shasum stub), asserting the fast path still installs a verified binary.
- ⚪ **[LOW]** \[security\] SHA-256 check is corruption-only, not a tamper boundary — binary and checksum share one origin/channel — `scripts/fetch-or-build.sh:96` · 2/4 models _(security)_
  The install now downloads and executes a remote prebuilt binary (previously it compiled cloned source). Integrity rests on SHA256SUMS fetched over the same HTTPS connection from the same release as the binary, with no signature — so anyone able to alter the binary (compromised release, or a MITM that can rewrite one response) can rewrite the checksum too. The check meaningfully guards against truncated/corrupted downloads but is not a security boundary against tampering. Advisory: the trust root (the GitHub repo/account) is the same as for the source build, so this is not a new exploitable vuln, just a narrower assurance than 'verified SHA-256' implies.
  _Fix:_ Either keep as-is and reword the success message/README to claim integrity (corruption protection) rather than authenticity, or publish a detached signature (cosign/minisign/gpg) and verify it before trusting SHA256SUMS to make the check a real tamper boundary.
- ⚪ **[LOW]** \[maintainability\] test temp directories can leak on panic — `tests/fetch\_or\_build.rs:155` · 1/4 models _(maintainability)_
  \`run()\` deletes its temp directory only at the normal return path. If an assertion panics before that line, the temp dir is left behind. Unique names prevent collisions, but the temp dir accumulates garbage over repeated test failures.
  _Fix:_ Wrap the temp directory in a small RAII guard that calls \`remove\_dir\_all\` on Drop, or use a crate like \`tempfile\`, so cleanup runs even when a test panics.
- ⚪ **[LOW]** \[test-coverage\] The no-download-tool and no-sha256-tool fallback branches are not exercised — `tests/fetch\_or\_build.rs:88` · 1/4 models _(test-coverage)_
  The hermetic harness always provides a stubbed curl and the host's real sha256sum/shasum, so the \`download … return 127\` path (no curl AND no wget) and the \`sha256\_of … return 127\` path (no sha256sum AND no shasum) — both of which call fallback with a tool-availability message rather than an asset message — are never run. A regression that broke either branch (e.g. dropping the wget/shasum alternative, or mishandling the 127 exit) would not be caught.
  _Fix:_ Add a variant that runs with curl/wget absent from the stub PATH and one with sha256sum/shasum absent, asserting the script falls back and never installs an unverified binary.

- **[MEDIUM]** \[security\] Third-party actions referenced by mutable tags — `.github/workflows/release.yml:25`
  _Dismissed:_ Single-model (kimi lens-security) medium; advisory by calibration. Code-checked: the existing .github/workflows/ci.yml already pins every action by version tag (incl. dtolnay/rust-toolchain@master, looser than @1.96.0), so pinning only release.yml is an inconsistent half-measure. Owner explicitly decided 2026-06-23 to keep tag pins and logged a repo-wide SHA-pin as a follow-up in AUDIT.md. Release assets are SHA-256-verified at install. Not a merge-blocker.

smarzban:

I ran the full panel (holistic ×4: kimi, glm, opus, gpt-5.5 + lens-security ×2: gpt-5.5, kimi) plus the deterministic scan (0 findings). I read the code behind every gating cluster before deciding.

**Two real issues I'm fixing:**

1. **🟠 Version-keyed prebuilt can serve a stale binary (3 models).** The fast path fetches `v<Cargo.toml version>`, but this repo keeps `main` at the last-released version *between* releases (no `-dev` suffix). So a bare install of a `main` that has advanced past the tag would install the *older tagged* binary for *newer* source — a regression from today's always-build-from-source, and it makes the README's "builds from source when main is ahead" claim false. **Fix:** gate the fast path on the checkout being exactly the `v<version>` tag commit (via `git rev-parse`); otherwise fall back to a source build. This restores the design's approved invariant — *the prebuilt is used only when the source is exactly a released commit*.

2. **🟠 Build job has `contents: write` (2 models).** The top-level `permissions: contents: write` leaks to the build matrix, which only compiles + uploads artifacts. **Fix:** top-level `contents: read`, keep `write` scoped to the `publish` job (least privilege).

**Dismissed (with code-checked reasons):**
- **Mutable-tag actions (`release.yml:25`, single model):** consistent with the existing `ci.yml` (which tag-pins everything, incl. `dtolnay/rust-toolchain@master`); the owner explicitly chose to keep tag pins and logged a repo-wide SHA-pin as a follow-up. Pinning only `release.yml` would be an inconsistent half-measure.
- **Env-override trust (`:21`, low):** `FV_*` overrides are for hermetic tests; exploiting them requires already controlling the install process's environment.
- **Unsigned `SHA256SUMS`/TOFU (`:86`, low):** documented + accepted design decision (checksums now, signatures deferred — they layer on later with zero install-UX change).
- **Grep-fragility (`:63`, low) + coverage/temp-leak lows:** the `^version` anchor only matches the `[package]` line; the no-curl/no-sha branches can't be exercised hermetically without losing coreutils from PATH (their outcome is the safe fallback either way). Logged as follow-ups.

**Decision: 🔴 Request changes** — implementing the two fixes now; Round 2 will verify.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ tools + verify · models: deterministic, codex:gpt-5.5

### Progress since Round 1
✅ Resolved (3): \[security\] build job inherits unnecessary contents:write permission; \[correctness\] Version-keyed prebuilt serves a stale release binary when main is ahead, contradicting the documented fallback; \[correctness\] Prebuilt fetch can install a stale binary for newer source — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

Round 2 was single-model verification (codex gpt-5.5) of Round 1's three blocking findings against the fix commit `1e2c089`, plus a fresh deterministic scan (0). The verifier returned an empty array — all three resolved, no regression. **Verdict: ✅ PASS.**

### What this PR implements
A prebuilt-binary install path (v1.2.0): the `[[build]]` step runs `scripts/fetch-or-build.sh`, which downloads a SHA-256-verified binary matching the source's version + platform and falls back to `cargo build` on any miss. A `release.yml` workflow cross-builds the three targets on `v*` tags and publishes them + `SHA256SUMS`. Install command unchanged.

- **Stale-binary correctness (3 models):** the fast path now gates on the checkout being exactly the `v$version` tag commit (`git rev-parse`); a `main` that is ahead of its tag falls back to a source build, so the prebuilt is used only when the source is exactly a released commit. Restores the design's "binary always matches the cloned source" invariant; the README claim is now accurate. Two new tests cover at-tag → prebuilt and ahead-of-tag → fallback (no download).
- **Least-privilege CI (2 models):** top-level workflow permission dropped to `contents: read`; `contents: write` is scoped to the `publish` job only.

### Deferred / not covered (advisory, logged)
- Mutable-tag action pins — owner decided to keep tag pins (consistent with `ci.yml`); repo-wide SHA-pin logged in AUDIT.md.
- Unsigned `SHA256SUMS`/TOFU — accepted design decision (checksums now, signing layers on later).
- No-curl/no-sha and wget/shasum fallback branches untested (hermetically untestable without removing coreutils from PATH; outcome is the safe fallback); test temp-dir leak-on-panic — minor hygiene.
- GitHub Actions execution itself is verified by cutting a real `v1.2.0` tag (manual step).

### Decision: ✅ Approve
The gate verdict is `pass` on HEAD `1e2c089`. Merge + the v1.2.0 release are owner-gated — holding for the owner's go before merging and tagging.

PR #1, opened 2026-06-20
PR1 — Foundation & pure logic (T-1…T-5)

smarzban:
## What

First of five layer-based PRs executing `specs/file-viewer/plan.md`. Establishes the crate, the herdr manifest, and the four leaf/pure components, each built test-first (red-green-refactor), repo green between every task.

### Tasks
- **T-1** — Scaffold crate (`herdr_file_viewer` lib + `herdr-file-viewer` bin), techstack-pinned deps, and `herdr-plugin.toml` declaring a **split-pane launch** (AC-17) with **no event hooks** (AC-N4).
- **T-2** — **View Policy**: `default_mode` precedence (changed→Diff even for markdown, else markdown→Rendered, else Syntax) + `applicable_modes` cycle set always including RawContent override. (AC-8, AC-9, AC-10, AC-11)
- **T-3** — **Root Resolver** + `LaunchContext`: root = git toplevel in a repo (AC-1) else cwd (AC-2); not-a-repo is normal, not an error (AC-26); linked worktrees flagged.
- **T-4** — **Git Service** status: `git status --porcelain` → per-file `Status` map (AC-7), read-only (AC-N2).
- **T-5** — **Git Service** baseline/changed-set/diff: context-smart default baseline (AC-14/AC-15), toggling changes the set + diff (AC-16), raw unified diff text (AC-9), read-only (AC-N2).

## Tests
22 tests green: `manifest`, `root`, `git_status`, `git_baseline` (integration) + `view_policy` (unit). Release binary builds. Adds a dependency-free `tests/common` helper (TempDir + git runner) honoring the pinned dev-deps (no `tempfile`).

## Notes
- Confirms the flagged build-time pairings: ratatui 0.30.x + crossterm 0.29 + ansi-to-tui 8.0.1 + ignore + serde + insta + expectrl all resolve and compile together.
- Git Service paths are repo-root-relative (git-native); the base-branch hint is honored in `default_baseline`, with a `main`/`master` fallback inside `changed_set`/`diff` per the plan's signatures.

PR #38, opened 2026-06-23
feat: copy selected file's path to the clipboard (y / Y)

### LINKED ISSUE #37: Adding a yank key for file paths
Great tool! Thanks for sharing.

How about adding a key to copy to the clipboard the fullpath / repo path of the highlighted file? maybe Y and y ?

riclib:
Closes #37.

Adds two keyboard shortcuts to the viewer:

- **`y`** copies the selected file's **repo-relative** path (e.g. `src/app.rs`)
- **`Y`** copies its **absolute** path

Both surface a `Copied …` confirmation in the notices strip, work regardless of which pane is focused (and while zoomed), and also work for directories. Like every other key, they're **read-only** — they only copy a path string, never touching file contents (AC-N3).

### How it works
The copy uses the terminal's **OSC 52** clipboard escape — the portable, dependency-free way for a pane to set the *host* terminal's clipboard. It travels through herdr (and SSH) with no extra tooling, unlike a native clipboard API bound to the local display. A confirmation toast from the terminal/multiplexer is a nice bonus where supported.

The OSC 52 payload is base64-encoded with a small in-crate RFC 4648 encoder (**no new dependency**, matching the project's minimal-deps style). Base64 also confines an attacker-chosen file name (trust boundary #1) to `[A-Za-z0-9+/=]`, so a name containing `ESC`/`BEL` can't break out of the escape sequence and drive the terminal — the same defense-in-depth the renderer path applies to file content.

### Design
Follows the existing trait-injection pattern so the controller stays unit-testable:
- new `Clipboard` trait in `Components` (live OSC 52 impl in `app`, recording stub in tests)
- two new read-only `Intent`s (`CopyRepoPath` / `CopyAbsPath`), wired through `input` with `Y` riding the existing "Shift permitted" rule
- a single `copy_path` handler in the controller

### Tests
- `input`: `y`/`Y` map to distinct intents; a Ctrl chord on the key fires neither
- `controller`: `y` copies the repo-relative path + confirms; `Y` copies the absolute path
- `app`: base64 matches RFC 4648 vectors (incl. padding); control bytes stay within the safe alphabet

`cargo build`, `cargo clippy --all-targets`, and `cargo fmt --check` are clean. All tests pass except one pre-existing, unrelated failure in `update::tests::hardened_git_ignores_a_malicious_repo_local_insteadof` (fails on a clean checkout too — an environment/git-config issue, untouched by this change).

### Notes for the maintainer
- The CHANGELOG entry sits under `[Unreleased]`; I didn't bump `Cargo.toml` / `herdr-plugin.toml` versions — that's your release call (this is a minor feature → `1.3.0`).
- No new runtime dependencies.

https://claude.ai/code/session_01UygKVzkcVg2cM5dZWQiS78

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -43,7 +43,7 @@ herdr → env (HERDR_PLUGIN_CONTEXT_JSON)
           │
    host::from_env → root::resolve → git::default_baseline
           │
-   Controller::new  ── wires live GitService / ContentProvider / EditorHandoff behind traits
+   Controller::new  ── wires live GitService / ContentProvider / EditorHandoff / Clipboard behind traits
           │
    event loop (app::run):  draw → poll input → handle(intent) → drain finished renders → repeat
 ```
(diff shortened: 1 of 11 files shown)


smarzban:
Pushed a small hardening commit, strips control bytes from the copied path. Thanks for the PR!

Issue #40, opened 2026-06-24
Browse worktrees


mitralone:
Can we have the option to browse the worktree the agent is working on? Right now the browse pane defaults to project base. Or the ability to change folders is also fine.

smarzban:
Hey @mitralone, this is in as of v1.4.0. Press W to open the worktree picker and switch the viewer to another worktree of the repo; it'll preselect whichever one your agent is working in. Re-run `herdr plugin install smarzban/herdr-file-viewer` to grab it. Give it a go when you get a chance and let me know if it works the way you had in mind, or if anything feels off — happy to tweak it.

mitralone:
Hey @smarzban works great, thanks. 

But this time I seem to have another problem with this version. If I have files more than what fits into visible area, I can not scroll down. The cursor moves out of the screen.

smarzban:
Fixed in v1.5.0, the tree scrolls with the selection now, and I added scrollbars while I was in there (draggable, too). Re-run `herdr plugin install smarzban/herdr-file-viewer` to grab it. Let me know if it's behaving for you, thanks for flagging it!

mitralone:
Works great. Good job!

PR #41, opened 2026-06-25
feat: switch worktree — re-root the viewer to another git worktree (SMA-63)

smarzban:
## Switch worktree (re-root the tree)

A keyboard-driven **worktree picker** that **re-roots the running viewer in place** to another git
worktree of the current repository — no relaunch, same session. Addresses the external request in
#40 (tracked as SMA-63).

Press **`W`** → a centered picker lists the repo's worktrees (path + branch, or a detached marker) →
arrows/`j`/`k` to move, **Enter** to switch, **Esc** to cancel. The tree, git status/diffs, and
content pane rebuild for the chosen worktree; the viewer never opens a new pane and never mutates
anything.

### What it does
- Lists the current repo's worktrees via read-only `git worktree list --porcelain -z` (bare excluded).
- When herdr reports an agent-hosting worktree (via `$HERDR_BIN_PATH` `worktree list`/`agent list`),
  the picker **pre-selects** it; otherwise it pre-selects the current root. Degrades cleanly to a
  git-only picker when herdr metadata is unavailable.
- Confirming re-roots **in place**: re-resolve → rebuild Tree + Git Service + Content Renderer →
  respawn the render worker. Structural re-root is synchronous (interactive within the launch
  budget); the heavier git status/changed-set fills in **asynchronously**, so input is never blocked.
- **Preserves** view prefs (tree width, wrap, changed-only, show-ignored, baseline mode); **resets**
  navigation (cursor to root, expansions cleared, picker closed, focus on tree, unzoomed).

### Read-only & scope (negative criteria)
- No file or git mutation; only read-only queries (no checkout, no branch change). Re-roots to a
  worktree **directory** — it never checks out a branch.
- No persistent state — a relaunch re-resolves the root normally.
- Picker offers only the current repo's worktrees (no arbitrary dir, no other repo).
- Re-roots **only** on the explicit `W` action — no herdr event or automatic trigger; the manifest
  declares no `[[events]]`.

### Architecture
Three new components (Worktree Provider, Re-root Engine, Worktree Picker) + one construction seam,
fitting the existing single-in-process-TUI design.
- **ADR-0003 — runtime re-rooting:** the viewer resolved its root once at launch; this adds a
  user-initiated runtime re-root (like Refresh, never automatic).
- **ADR-0004 — Provider Factory:** `Controller::new` now takes a `Resolved -> {git, content}` factory
  instead of fixed provider instances, so a re-root can rebuild the root-bound services and respawn
  the render worker while keeping tests hermetic.

### Acceptance criteria
All 22 criteria covered: **AC-1…AC-17** (picker, re-root, state-on-switch, degradation, performance)
and the five negative invariants **AC-N1…AC-N5** (read-only, read-only-git, no-persist, repo-only,
no-events). Each is exercised by tests (the keyboard operability of the picker, AC-5, by a pty e2e),
and AC-N5 additionally by a no-events manifest check + a gating test that no intent re-roots except
the explicit picker-confirm path.

### Quality
- **Zero new dependencies** (std + the already-pinned stack: ratatui 0.30.1, crossterm 0.29.0,
  serde/serde_json, system git CLI).
- **303 tests passing**, 0 failing. The async re-root and pty picker e2e were stress-run (reroot
  15×, e2e 10×) with zero flakes. `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check`
  both clean. Release build clean.

Built test-first, one reviewed atomic commit per plan task (T-1…T-17).

---
Addresses the request in #40. (PR/merge and any release tag are owner-gated — opened for review +
manual testing, not auto-merged.)


smarzban:
## Review Gate — Round 1
🚫 **BLOCK** — 5 blocking finding(s) must be resolved or justified.

Findings: 12 total — 0 critical · 0 high · 5 medium · 7 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-spec + lens-subtle-correctness + tools · models: deterministic, ollama:kimi-k2.7-code:cloud, ollama:glm-5.2:cloud, claude:claude-opus-4-8, codex:gpt-5.5

### Must fix
- 🟠 **[MEDIUM]** \[concurrency\] Pending re-root status can overwrite newer git state — `src/controller.rs:1353` · 2/4 models _(concurrency)_
  The status worker captures the baseline at re-root time, but \`poll\` unconditionally assigns its \`changed\` set later. If the user toggles baseline or performs a refresh/focus-gain while that slow worker is pending, the stale async result can overwrite the newer changed-set and leave changed-only filtering inconsistent with the current baseline.
  _Fix:_ Tag status requests with a generation plus root/baseline and discard mismatches in \`poll\`; also invalidate the pending status request whenever synchronous git state is refreshed or the baseline changes.
- 🟠 **[MEDIUM]** \[security\] Unsanitized worktree path reaches terminal notice — `src/controller.rs:373` · 1/4 models _(security)_
  A worktree path from \`git worktree list\` is untrusted and can contain terminal control bytes; if the selected worktree is missing or inaccessible, that raw path is interpolated into \`action\_notice\`, and notices are rendered without the \`sanitize\_label\` path used for picker rows. A stale/prunable worktree with an ESC/OSC-bearing path can therefore spoof or drive the terminal when the user activates it.
  _Fix:_ Sanitize control characters before storing the notice, or sanitize every notice at the presenter sink before rendering.
- 🟠 **[MEDIUM]** \[correctness\] Re-root drops the base branch hint — `src/controller.rs:379` · 2/4 models _(correctness)_
  \`re\_root\` resolves the target with \`LaunchContext::default()\`, so \`resolved.base\_branch\` is always lost before the provider factory builds the new \`LiveGit\`. In non-main/master repos the existing baseline tests show that losing this hint makes \`Baseline::Base\` degrade to HEAD and miss committed feature work after a worktree switch.
  _Fix:_ Carry the current base-branch hint through re-root, or attach the selected worktree's base branch from the herdr worktree row, and add a regression test switching in a trunk-based repo while using \`Baseline::Base\`.
- 🟠 **[MEDIUM]** \[correctness\] mouse events bypass the worktree picker modal — `src/controller.rs:736` · 3/4 models _(correctness)_
  Keyboard intents are correctly routed through \`handle\_picker\_intent\` while the picker is open, but \`handle\_mouse\` does not check \`self.picker\`. Clicks and wheel events therefore hit-test against the underlying two-column layout and can change the tree selection, content scroll, or focus behind the modal overlay. This creates a confusing state where closing the picker leaves the viewer on a different file/scroll position than when the picker opened.
  _Fix:_ Add an early guard at the top of \`handle\_mouse\` that returns \`Effects::noop()\` whenever \`self.picker.is\_some()\`, making mouse input inert while the modal picker is open (matching the keyboard modal behavior).
- 🟠 **[MEDIUM]** \[correctness\] worktree is\_current flag uses asymmetric path canonicalization — `src/worktree.rs:252` · 2/4 models _(correctness)_
  \`worktree::list\` canonicalizes \`current\_root\` before passing it to the parser, but \`parse\_record\` compares that canonical path against the raw git-emitted worktree path without canonicalizing it. On systems where paths traverse symlinks (e.g. macOS \`/tmp\` → \`/private/tmp\`) or when the caller reaches the repo through a symlink, the current worktree will not be marked \`is\_current\`, breaking the picker preselect (AC-4) and any test asserting exactly one current row. The code's own comment claims the comparison is 'symlink-stable'.
  _Fix:_ Canonicalize each worktree \`path\` in \`parse\_record\` with the same \`canonicalize().unwrap\_or\_else(\|\_\| path.clone())\` fallback, then compare the two canonicalized paths while still storing the original git path in \`Worktree::path\`.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[performance\] Picker open runs git + two herdr subprocesses synchronously on the input thread with no timeout — `src/controller.rs:1226` · 1/4 models _(performance)_
  open\_worktree\_picker shells \`git worktree list\` plus two \`herdr\` calls (worktree list --json, agent list --json) via Command::output() on the input thread. A hung or slow herdr binary freezes the UI until it returns; there is no timeout, unlike the bounded capture\_stdout used for diffs.
  _Fix:_ Bound the herdr/git calls (kill after a deadline, as capture\_stdout does for git diff) or move the enumeration off the input thread and populate the picker when it arrives, degrading to git-only on timeout.
- ⚪ **[LOW]** \[error-handling\] Status-refresh thread is not panic-contained; a panic leaves changed\_only never applied — `src/controller.rs:443` · 1/4 models _(error-handling)_
  Unlike the render worker (catch\_unwind), the dispatch\_status\_refresh thread has no unwind guard. If git.status()/changed\_set() panics, the sender is dropped silently, poll sees Disconnected and clears status\_rx, and set\_changed\_only is never called on the new tree — so the controller believes changed\_only is on while the tree shows everything.
  _Fix:_ Wrap the thread body in catch\_unwind (as spawn\_worker does) and on panic either apply the empty filter explicitly or surface a non-fatal notice so controller/tree state stays consistent.
- ⚪ **[LOW]** \[security\] Herdr subprocess output is unbounded and synchronous — `src/herdr.rs:83` · 1/4 models _(security)_
  \`run\_json\` uses \`Command::output()\` with no timeout or stdout cap on an external \`herdr\` command, so a hostile or wedged command can block the UI thread or exhaust memory when the picker opens. This crosses a process boundary and treats the entire stdout as trusted-size input.
  _Fix:_ Spawn the process with piped stdout, read up to a fixed JSON size limit, enforce a short timeout, kill/reap on overrun, and return an error so the picker degrades to git-only mode.
- ⚪ **[LOW]** \[correctness\] Picker overlay has no scroll; cursor highlight disappears below the popup — `src/presenter.rs:372` · 2/4 models _(correctness)_
  draw\_picker\_overlay renders all rows into a fixed 60%-height Paragraph with no scroll offset. When a repo has more worktrees than fit (\~11 on a 24-row terminal), NavDown moves the cursor past the visible rows and the highlighted row is rendered off-screen, so the user can't see what Enter will confirm.
  _Fix:_ Give the picker a scroll offset that follows the cursor (clamp the cursor into view, or use a List widget with paragraph scroll), mirroring how the tree keeps its selection visible.
- ⚪ **[LOW]** \[test-coverage\] AC-1 'opens regardless of how many worktrees' untested for the single-worktree repo — `tests/controller.rs:253` · 1/4 models _(test-coverage)_
  AC-1 emphasizes the picker opens whenever rooted in a git repo, 'regardless of how many worktrees exist'. Every picker-open test (switch\_worktree\_in\_a\_repo\_opens\_picker\_preselecting\_current, etc.) first adds a linked worktree so rows.len() \>= 2. No test opens the picker in a repo containing only the main working tree, so a regression adding a 'rows.len() \< 2 → no picker' guard would pass all tests.
  _Fix:_ Add a test that opens the picker in a freshly-initialized repo with no linked worktree and asserts the picker opens with exactly one (current) row.
- ⚪ **[LOW]** \[test-coverage\] AC-17 perf budget (1s on 10k files) is not exercised for re-root — `tests/reroot.rs:318` · 1/4 models _(test-coverage)_
  AC-17 is mapped 'integration / perf' and quantifies a 1-second navigable budget on a 10,000-file repo after a switch. re\_root\_status\_fills\_in\_asynchronously proves the git work is off-thread, and tree\_perf.rs times only the launch path (AC-22) via TreeModel::new directly. No test times the re\_root synchronous path on a large repo, so a regression that made the synchronous portion of a switch slow on 10k files would not fail any test.
  _Fix:_ Add a perf test that builds a \~10k-file repo, calls re\_root, and asserts the call returns and the tree is navigable within 1s (mirroring tree\_perf.rs but through the re\_root path).
- ⚪ **[LOW]** \[test-coverage\] AC-9 'diffs reflect the chosen worktree' is not asserted post-switch — `tests/reroot.rs:583` · 1/4 models _(test-coverage)_
  AC-9 requires both status indicators AND diffs to reflect the chosen worktree after a switch. Status markers are well covered (re\_root\_markers\_reflect\_new\_root\_git), but the diff half is not: re\_root\_render\_resolves\_through\_the\_respawned\_worker uses an empty diff and a FakeContent that ignores raw\_diff, and CannedGit.diff is never asserted to flow into a Diff-mode render after a re-root. A regression where the diff query bound to the old root would not be caught.
  _Fix:_ After a re-root, select a file in Diff/FullDiff mode and assert the rendered content carries the new root's git.diff output (e.g. a content provider that echoes raw\_diff, with CannedGit returning a B-specific diff string).

smarzban:

I ran the full panel (deterministic scan + holistic ×4: kimi-k2.7-code, glm-5.2, opus-4-8, gpt-5.5) plus three warranted lenses (security, spec, subtle-correctness). Every gating cluster was read against the code before adjudicating. The panel did real work — it surfaced cross-cutting integration issues that per-task review couldn't see. **Decision: 🔴 Request changes.** Fixing all five blocking mediums + the worthwhile lows this round:

**Blocking — fixing:**
1. **Stale async status can clobber newer git state / wrong view-mode after switch** (`controller.rs`, poll + `dispatch_status_refresh`). Two real bugs in one cluster: (a) `poll` unconditionally applies the re-root status worker's result, so toggling baseline / refreshing while it's pending overwrites the newer changed-set; (b) `re_root`'s synchronous `dispatch_render` runs against the just-cleared changed-set, and `poll` never re-dispatches the render — so a changed file shows in content/markdown mode instead of **Diff** after a switch (AC-9). Fix: a generation guard that discards stale status (and is invalidated by any synchronous refresh), plus a re-dispatch of the render when the fresh changed-set lands.
2. **Mouse bypasses the modal picker** (`controller.rs::handle_mouse`) — corroborated by 4 reviewer passes. `handle()` is modal while the picker is open, but `handle_mouse` has no guard, so a click/wheel behind the overlay still moves the tree cursor, scrolls, or even activates a node — and undermines AC-6 (cancel-unchanged) if the mouse was used. Fix: gate `handle_mouse` when the picker is open.
3. **`is_current` asymmetric path canonicalization** (`worktree.rs::list`) — `current_root` is canonicalized but compared against the raw git-emitted worktree path, so a symlinked / `/tmp`-style path mismatch flags no current row and the picker preselects row 0 instead of the current worktree (AC-4). Fix: canonicalize both sides.
4. **Re-root drops the base-branch hint** (`controller.rs::re_root`) — `LaunchContext::default()` loses `base_branch`, so after a switch the Base-mode diff baseline falls back to main/master auto-detection (wrong for a worktree forked from a non-default branch — AC-9). Fix: carry the launch base-branch hint into `re_root` (it's a repo-wide ref, re-validated against the target).
5. **Worktree path in the failure notice isn't sanitized** (`controller.rs`, AC-16 notice). Display-inert in practice (ratatui's renderer strips control graphemes, the project's documented AC-27 posture), but inconsistent — picker rows are `sanitize_label`'d and notices aren't. Fix: sanitize at the notice sink for a uniform guarantee.

**Also fixing (lows worth closing):**
- Picker overlay has no scroll — the cursor leaves the popup on repos with many worktrees (exactly herdr's multi-agent use case). Scroll the list to keep the selection visible.
- Status-refresh thread isn't panic-contained (the render worker is). Wrap it.
- Test coverage: add the missing AC-1 (single-worktree-repo picker opens), AC-17 (re-root perf budget on a large repo), and AC-9 (diff — not just status markers — reflects the new worktree post-switch) tests. The AC-9 test also guards bug B above.

**Deferred (advisory, not blocking) — for a follow-up:**
- Picker open runs git + two herdr queries synchronously on the input thread with no timeout/output cap (`controller.rs` / `herdr.rs`). herdr is trusted host infrastructure and git is fast, so this is a responsiveness/robustness nicety rather than a merge blocker; a bounded-timeout `run_json` is a reasonable later hardening.

Implementing these now; round 2 will verify each at the new HEAD.

smarzban:
## Review Gate — Round 2

Findings: 1 total — 0 critical · 0 high · 1 medium · 0 low · 0 info.
_Reviewed by:_ tools + verify2 · models: deterministic, claude:claude-opus-4-8

### Progress since Round 1
✅ Resolved (5): \[concurrency\] Pending re-root status can overwrite newer git state; \[security\] Unsanitized worktree path reaches terminal notice; \[correctness\] Re-root drops the base branch hint; \[correctness\] mouse events bypass the worktree picker modal; \[correctness\] worktree is\_current flag uses asymmetric path canonicalization — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (1): Re-root async status still clobbers a fast baseline-toggle (status\_rx not cleared in toggle\_baseline)

### Must fix
- 🟠 **[MEDIUM]** Re-root async status still clobbers a fast baseline-toggle (status\_rx not cleared in toggle\_baseline) — `src/controller.rs:1080` · 1/1 models _(concurrency)_
  The R1 fix drops the pending re-root async fetch (status\_rx = None) only inside refresh\_git\_state, covering the r-refresh / focus-gain / editor-return paths. But the original finding explicitly named 'toggles baseline', and toggle\_baseline (controller.rs:1070) does NOT call refresh\_git\_state and does NOT clear status\_rx. After a switch, dispatch\_status\_refresh leaves status\_rx = Some with the OLD baseline captured by value. Input is never blocked (AC-17), so the user can press 'b' before the async fetch lands; toggle\_baseline recomputes self.changed against the NEW baseline synchronously but leaves status\_rx pending. On the next poll, the stale async result (status, changed) overwrites self.changed with the pre-toggle baseline's set, re-applies the changed-only filter against it, and — via the new AC-9 dispatch\_render() — re-renders using descriptor().is\_changed off the stale set, so a file can render in the wrong mode (Diff vs content) with markers reflecting the wrong baseline. The fix's own test comment lists 'baseline-toggle' as a covered sync path, but the test only drives Intent::Refresh, so the gap is untested. Self-heals on the next refresh/toggle, matching the original medium severity.
  _Fix:_ Clear the pending re-root fetch in the baseline path too: add self.status\_rx = None to toggle\_baseline after it recomputes self.changed (it has produced the authoritative changed-set for the new baseline). Equivalently, route toggle\_baseline's recompute through refresh\_git\_state, or factor a small drop\_pending\_status() helper used by both. Then extend the existing test to also assert it with Intent::ToggleBaseline.

smarzban:
## Orchestrator review — Round 2 (verification)

Single-model verification (opus) of the five round-1 blocking findings at the fixed HEAD. **Four of five are fully resolved** (mouse-modal, is_current canonicalization, base-branch carry, notice sanitization — all confirmed at HEAD, no regressions). **Decision: 🔴 Request changes** — one prior finding is only *partially* resolved:

- **Stale re-root async status — the baseline-toggle path is still racy.** The R1 fix dropped the pending async fetch in `refresh_git_state` (covering refresh / focus-gain / editor-return), but `toggle_baseline` recomputes `self.changed` directly without going through `refresh_git_state` and never clears `status_rx`. The original finding explicitly named "toggles baseline," so this path is unfixed: after a switch, pressing `b` before the async status lands recomputes against the new baseline, then the stale async result clobbers it on the next `poll` (now also re-rendering off the stale set via the AC-9 fix). The R1 fix comment even listed baseline-toggle as covered, but the test only drove `Intent::Refresh`, so the gap was untested.

Verified in the code (`toggle_baseline` at controller.rs:1070 — no `refresh_git_state`, no `status_rx` clear). Fixing now: clear the pending async fetch in the baseline path (mirroring `refresh_git_state`) and extend the staleness test to also exercise `Intent::ToggleBaseline`. Round 3 will verify.

smarzban:
## Review Gate — Round 3

_Reviewed by:_ tools + verify3 · models: deterministic, codex:gpt-5.5

### Progress since Round 2
✅ Resolved (1): Re-root async status still clobbers a fast baseline-toggle (status\_rx not cleared in toggle\_baseline) — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:
## Orchestrator review — Round 3 (final) ✅

Round-3 verification (gpt-5.5, independent of round 2's opus) confirms the last finding resolved with no regression. The gate converged: **round 1** ran the full panel (deterministic scan + holistic ×4: kimi-k2.7-code, glm-5.2, opus-4-8, gpt-5.5 + lens-security / lens-spec / lens-subtle-correctness) and found real cross-cutting issues; **round 2** (opus) caught an incomplete fix; **round 3** (gpt-5.5) is clean. Every gating finding was read in the code before adjudication.

### What this PR implements
A keyboard worktree picker (`W`) that re-roots the viewer in place to another git worktree — read-only, async-status-filled, prefs preserved / nav reset, agent-aware pre-selection with git-only fallback. 22 ACs, ADR-0003/0004, 313 tests.

- **Re-root async-status race + view-mode (AC-9):** a pending re-root status fetch could clobber newer git state; the post-switch render ran against an empty changed-set so a changed file showed in the wrong (non-Diff) mode. Now: a shared `drop_pending_status()` invalidates the pending fetch on *every* synchronous recompute (`refresh_git_state` **and** `toggle_baseline` — the round-2 gap), and `poll` re-dispatches the render when the fresh changed-set lands. Guarded by Refresh- and ToggleBaseline-staleness tests + an AC-9 diff-mode-after-switch test.
- **Mouse bypassed the modal picker (AC-6):** `handle_mouse` is now inert while the picker is open. Guarded by a test.
- **`is_current` asymmetric path canonicalization (AC-4):** `list` now canonicalizes both sides; guarded by a `#[cfg(unix)]` symlink test.
- **Re-root dropped the base-branch hint (AC-9):** the launch hint is carried into `re_root` (recovers the common shared-base case; falls back to auto-detect otherwise).
- **Notice sanitization (AC-27):** notices now go through `sanitize_label` at the sink (uniform with rows/title — display was already inert via ratatui, now explicit).
- **Picker overlay scroll:** the cursor stays visible on many-worktree repos (herdr's use case). Status-refresh thread is now panic-contained.
- **Coverage added:** AC-1 single-worktree picker, AC-17 re-root perf budget, AC-9 post-switch diff.

### Deferred / not covered (non-blocking)
- **Synchronous herdr/git on picker-open with no timeout/output cap** (`herdr.rs` / picker open) — herdr is trusted host infrastructure and git is fast, so this is a responsiveness/robustness follow-up, not a merge blocker; a bounded-timeout `run_json` is a reasonable later hardening.
- **Base-branch on re-root is best-effort** — recovers the launch hint (shared-base worktrees) and otherwise auto-detects main/master; a future enhancement could query the target worktree's own base if herdr exposes it.

**Decision: ✅ Approve** (agrees with the `pass` verdict). Held for the owner's manual testing + merge — not auto-merged.

smarzban:
## Post-review fix (found during live owner-testing)

Commit `24be7b3` — added **after** the round-3 `pass` (which was against `b5403cc`), surfaced by running the real binary against live herdr 0.7.x:

**Bug:** `agent_preselect` queried `herdr agent list --json`, but this herdr's `agent list` **rejects `--json`** (exits non-zero; it prints JSON by default). So the query errored → `agent_active` returned `None` → the **AC-3 agent-active pre-selection silently never fired live**, always falling back to the current root. The 313 tests + the 4-model gate missed it because every test fed *canned* JSON and none exercised the actual herdr CLI contract — the documented "herdr JSON drift" risk, materialized.

**Fix:** drop `--json` from both the `worktree list` and `agent list` queries (both print JSON by default on 0.7.x; the worktree query tolerated the flag, the agent query didn't). Pinned the exact argv in the `full_switch_…` spy test as a regression guard. Whole suite green (313), clippy `-D warnings` + fmt clean.

**Verified live:** opened the picker against real herdr with an agent running in a second worktree — the overlay query now runs without error and the picker lists all worktrees. (This 2-line argv fix + regression test was verified live rather than re-running the full gate.)

smarzban:
## Merging ✅

Cumulative state since the round-3 pass:
- **Two herdr-CLI contract fixes** (found via live testing against herdr 0.7.x): dropped `--json` (`agent list` rejects it) and parse the `result.{worktrees,agents}` envelope — the agent overlay was silently dead live; now verified working (agent-status badges render against a real agent).
- **Picker enhancements**: per-row current marker + agent-status badge, content-sized/responsive box, blue (ANSI) border, horizontal scroll, herdr-style `esc close` + footer key-hints, uniform inner padding.
- **Hardening from a focused re-gate**: a fail-on-mutation guard test for the real-agent filter, hscroll clamp fixed to row width, doc corrections; one dismissed false-positive (a reviewer's own uncommitted mutation-test edit — committed code was always correct).
- **macOS CI fix**: a reroot test compared a guarded (non-rebuilt) raw root against a canonicalized path — broke only on macOS (`/var/folders`→`/private`). Fixed.

All 6 CI checks green (ubuntu + macOS × {1.96, stable}, clippy/fmt, audit); a single-model re-verify on the clean tree came back with 0 findings. Deferred follow-ups (non-blocking): bounded-timeout for the synchronous herdr query on picker-open; byte-oriented porcelain parsing for non-UTF8 worktree paths.

Merging via merge commit.

PR #47, opened 2026-06-25
fix(tree): scroll to keep selection visible + overflow scrollbars (#45)

### LINKED ISSUE #45: Tree scroll is not scrolling properly
When there is a large list of items, the view doesn't scroll, I can see the files/folders being selected but the left tree doesn't move.

Image

smarzban:
### The bug (#45, reported by @julianduque)
With more files than fit the pane, the tree stayed pinned to the top: the selection highlight kept moving but the viewport never followed, so once the cursor passed the last visible row you were selecting files you couldn't see.

**Root cause:** `draw_tree` rendered every node into a `Paragraph` with no scroll offset (always anchored at row 0). Rows past the interior height were simply clipped. The picker overlay already solved exactly this — the tree just never adopted it.

### The fix
- **Tree follows the selection.** `draw_tree` now applies a vertical scroll offset via the shared `scroll_offset(cursor, len, visible)` helper (generalized from the picker's `picker_scroll_offset`), so the selected row is always on screen.
- **Clicks stay correct when scrolled.** The offset is fed back to the controller via a new `PaneGeometry.tree_scroll`, and `hit_test` adds it — so a mouse click maps to the row *actually drawn* there. `geometry()` and `draw_tree` derive the offset from the same pure function, preserving the "drawn layout and hit-test geometry never disagree" invariant.

### Scrollbars (requested in review)
Scrollbars now appear **only where there is something to scroll** (`ratatui::widgets::Scrollbar` — already a transitive dep, no new Cargo dependency):
- **Tree** — a vertical bar when there are more nodes than fit.
- **Content** — a vertical bar when the file is taller than the pane, and a horizontal bar when an unwrapped line is wider than the pane.

They sit on the pane borders (track + thumb, no arrow glyphs) and are a no-op when everything fits. The picker overlay is intentionally left as-is (it already keeps its cursor in view and shows a key-hint footer).

### Tests
Test-first (red→green) throughout. New tests:
- `tree_scrolls_to_keep_selection_visible` — the reported bug, scoped to the tree interior.
- `geometry_reports_the_tree_scroll_offset_for_hit_testing` + `a_tree_click_maps_through_the_scroll_offset` — the click-mapping coupling.
- `tree_shows_a_vertical_scrollbar_only_when_it_overflows`, `content_pane_shows_a_vertical_scrollbar_when_content_overflows`, `content_pane_shows_a_horizontal_scrollbar_for_a_too_wide_unwrapped_line`.

Full suite **341 passing**; `cargo fmt --check`, `clippy -D warnings`, and `cargo build --release` all clean.

### Docs
CHANGELOG (`Unreleased` → `Fixed` + `Added`), README feature list, and ARCHITECTURE presenter row all updated in this PR.

Closes #45

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -31,7 +31,7 @@ is unit-testable with stubs.
 | `tree` | The rooted, `.gitignore`-aware file tree: filters, cursor, expansion, status markers. |
 | `view_policy` | A pure decision: which view mode a file gets (changed → diff, markdown → rendered, else → syntax content) and the cycle order. |
 | `render` | Produce the content-pane text: classify the file, delegate styling to an external CLI, and **neutralize escape sequences** before display. |
-| `presenter` | Draw the two-column (or zoomed / narrow) layout with ratatui; report the content viewport + pane geometry back. |
+| `presenter` | Draw the two-column (or zoomed / narrow) layout with ratatui; scroll the tree to keep the selection in view and draw scrollbars where a pane overflows; report the content viewport + tree scroll offset + pane geometry back (the controller adds the offset when hit-testing a tree click). |
 | `picker` | The modal worktree-switcher overlay state (rows, cursor, horizontal scroll) drawn over the layout; captures its own nav / confirm / cancel keys while open. |
 | `input` | Map crossterm key events → intents. |
 | `intent` | The closed set of user intents (one exhaustive enum). |
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -31,11 +31,11 @@ is unit-testable with stubs.
 | `tree` | The rooted, `.gitignore`-aware file tree: filters, cursor, expansion, status markers. |
 | `view_policy` | A pure decision: which view mode a file gets (changed → diff, markdown → rendered, else → syntax content) and the cycle order. |
 | `render` | Produce the content-pane text: classify the file, delegate styling to an external CLI, and **neutralize escape sequences** before display. |
-| `presenter` | Draw the two-column (or zoomed / narrow) layout with ratatui; scroll the tree to keep the selection in view and draw scrollbars where a pane overflows; report the content viewport + tree scroll offset + pane geometry back (the controller adds the offset when hit-testing a tree click). |
+| `presenter` | Draw the two-column (or zoomed / narrow) layout with ratatui; scroll the tree to keep the selection in view (and horizontally for long rows) and draw scrollbars where a pane overflows; report the content viewport + tree scroll offset + widest tree row + pane geometry back, so the controller can hit-test a tree click and map a scrollbar drag. |
 | `picker` | The modal worktree-switcher overlay state (rows, cursor, horizontal scroll) drawn over the layout; captures its own nav / confirm / cancel keys while open. |
 | `input` | Map crossterm key events → intents. |
 | `intent` | The closed set of user intents (one exhaustive enum). |
-| `controller` | Orchestrate intents → state changes; hold the ephemeral session state; dispatch renders to the worker; on a worktree switch, rebuild the root-bound services through a provider factory and respawn the render worker. |
+| `controller` | Orchestrate intents → state changes; hold the ephemeral session state; dispatch renders to the worker; map mouse events (clicks, wheel, divider + scrollbar drags) against the fed-back geometry; on a worktree switch, rebuild the root-bound services through a provider factory and respawn the render worker. |
 | `editor` | Hand a file off to `$EDITOR` (launch only — never reads or writes the file). |
 | `launch` | The "launch-or-focus-or-toggle" decision behind the shell launch scripts (pure, hermetically testable). |
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to this project are documented here. The format is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **The tree now scrolls to follow the selection.** On a long file list the tree stayed pinned to
+  the top, so moving the cursor past the last visible row selected files you couldn't see. The tree
+  now scrolls to keep the selected row in view (mouse clicks still map to the right row when
+  scrolled). ([#45](https://github.com/smarzban/herdr-file-viewer/issues/45))
+
+### Added
+- **Scrollbars** appear on the tree and content panes whenever there's more to see than fits — a
+  vertical bar when the list or file is taller than the pane, and a horizontal bar when an unwrapped
+  line is wider than the content pane. They show only where there is something to scroll.
+
 ## [1.4.0] - 2026-06-25
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,13 @@ All notable changes to this project are documented here. The format is based on
 
 ### Added
 - **Scrollbars** appear on the tree and content panes whenever there's more to see than fits — a
-  vertical bar when the list or file is taller than the pane, and a horizontal bar when an unwrapped
-  line is wider than the content pane. They show only where there is something to scroll.
+  vertical bar when the list or file is taller than the pane, and a horizontal bar when a row /
+  unwrapped line is wider than the pane. They show only where there is something to scroll, and the
+  bars are **draggable with the mouse** (drag ↕ to scroll vertically, ↔ to scroll horizontally;
+  pressing the track jumps to that position).
+- **The tree scrolls horizontally** so a long or deeply-nested file name can be read in full — via
+  the horizontal mouse wheel or by dragging the tree's horizontal scrollbar (the `←`/`→` keys stay
+  expand/collapse in the tree).
 
 ## [1.4.0] - 2026-06-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,10 @@ All notable changes to this project are documented here. The format is based on
 ### Added
 - **Scrollbars** appear on the tree and content panes whenever there's more to see than fits — a
   vertical bar when the list or file is taller than the pane, and a horizontal bar when a row /
-  unwrapped line is wider than the pane. They show only where there is something to scroll, and the
-  bars are **draggable with the mouse** (drag ↕ to scroll vertically, ↔ to scroll horizontally;
-  pressing the track jumps to that position).
+  unwrapped line is wider than the pane. They show only where there is something to scroll. The
+  bars render **inside** the pane (one cell off the text) and are **draggable with the mouse**:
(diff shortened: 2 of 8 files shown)


smarzban:
## Review Gate — verdict: ✅ pass (after 1 fix round)

**Reviewed by** (holistic, round 1): `kimi-k2.7-code` · `glm-5.2` · `claude-opus-4-8` · `gpt-5.5` + deterministic scan. **Verified** (round 2): `gpt-5.5`. Full 4-lineage panel (opus's backend crashed once and was re-run to restore the 4th lineage).

### Round 1 — findings & disposition

| Sev | Finding | Models | Disposition |
|---|---|---|---|
| medium | Clicking a visible row in a *scrolled* tree jumped the viewport (and could make a double-click activate the wrong row) — the offset was a pure function of the cursor | codex | **Fixed** `e92b21e` — `sticky_scroll_offset` scrolls minimally from the last frame's offset; a row already in view doesn't move |
| medium | `offset as u16` wraps for >65535 rows above the fold | codex, glm, opus (×3) | **Fixed** `e92b21e` — saturating cast |
| low | Scrollbar thumb never reached the track end at max scroll | codex | **Fixed** `e92b21e` — `scrollbar_state` models the scroll range as `(max_scroll+1)` positions so the thumb reaches the bottom at the last offset (verified: top at offset 0, bottom at max) |
| low | Stale `// the tree does not scroll` comment in `scroll_at` | glm | **Fixed** `e92b21e` |
| low | Vertical content scrollbar can be missed under **wrap** (raw line count vs wrapped rows) | codex, opus (×2) | **Deferred** — the accurate count needs `rendered_line_count()`, deliberately kept off the per-frame path for perf; exact for the unwrapped case (code/diffs, the primary scroll use); wrapped prose stays key-scrollable. Documented limitation. |
| low | Content vscrollbar track spans the (0–2 row) notices strip, slightly mis-sizing the thumb when notices are present | opus | **Deferred** — cosmetic imprecision in a rare state (notices only show on truncation / renderer-fallback); the bar still reads correctly. |
| info | Tree vscrollbar shares the divider column | opus | **Won't fix** — the tree's right border *is* the column abutting content; conventional placement, and the divider drag is glyph-independent (hit-test unaffected). |

### Round 2 — verification (gpt-5.5)
Scoped to the round-1 findings + the fix diff: **clean (`[]`)** — every fix holds at HEAD, no regression. Deterministic scan clean both rounds.

### Verdict
The two mediums and the two clear-correctness lows are fixed and verified; the three remaining low/info items are advisory precision/cosmetic nits on the new scrollbar and are deferred with the reasons above (per gate calibration — lows are advisory unless they're real merge-blockers, which these aren't). **✅ pass.**

*Not merging — left for the owner to test + merge (owner-gated).*

smarzban:
## Review Gate — Round 1

Findings: 9 total — 0 critical · 0 high · 3 medium · 6 low · 0 info.
_Reviewed by:_ holistic + lens-simplify + tools · models: deterministic, kimi-k2.7-code:cloud, glm-5.2:cloud, claude-opus-4-8, gpt-5.5

### Advisory (non-blocking)
- ⚪ **[LOW]** \[maintainability\] Press and Drag handlers duplicate the 5-arm Drag→method match — `src/controller.rs:844` · 2/4 models _(maintainability)_
  The Down(Left) match and the Drag(Left) match each carry the same five \`Drag::X =\> self.scroll\_X\_to\_\*(col\|row)\` arms; adding a sixth draggable target means editing two match trees in lockstep, the kind of duplication that drifts.
  _Fix:_ Add \`fn apply\_drag(&mut self, drag: Drag, col: u16, row: u16) -\> Effects\` holding the single match; the Down arm sets \`self.drag\` then calls it (after the Divider/scrollbar press), the Drag arm calls it directly.
- ⚪ **[LOW]** \[correctness\] scrollbar drag does not clear last\_click, risking a false double-click — `src/controller.rs:865` · 1/4 models _(correctness)_
  The mouse-up handler returns early when a drag ends without clearing \`last\_click\`, so a prior tree-row click can still pair with a later click on the same screen row within the double-click window. After a scrollbar (or divider) drag scrolls the viewport, the same screen row may now map to a different node, causing an unintended activate/zoom.
  _Fix:_ Clear \`self.last\_click = None\` when a drag ends (in the \`MouseEventKind::Up\` branch after \`drag.take()\`), or whenever any drag starts, so drag interactions reset the double-click sequence.
- ⚪ **[LOW]** \[maintainability\] Four scroll\_\*\_to\_row/col methods are near-identical copy-paste — `src/controller.rs:995` · 1/4 models _(maintainability)_
  scroll\_content\_to\_row, scroll\_content\_h\_to\_col, and scroll\_tree\_h\_to\_col differ only in (track field, max source, target field) — same \`Some(track)? → max → track\_fraction → guard span==0\|\|max==0 → offset = round(rel\*max/span) → assign → redraw\` body, copied three times. The fourth (scroll\_tree\_to\_row) is genuinely different (sets the cursor) so it rightly stays separate.
  _Fix:_ Extract \`fn offset\_from\_track(track: Option\<Rect\>, pos: u16, horiz: bool, max: u16) -\> Option\<u16\>\` returning the rounded offset (or None when degenerate); each of the three methods becomes a 3-line \`let Some(off) = Self::offset\_from\_track(...); self.field = off; Effects::redraw()\`, keeping the per-field assignment at the call site.
- ⚪ **[LOW]** \[maintainability\] draw\_vscrollbar and draw\_hscrollbar differ only by orientation + thumb glyph — `src/presenter.rs:233` · 1/4 models _(maintainability)_
  The two functions are byte-for-byte parallel: same overflow guard, same scrollbar\_state call, same begin/end/track(None) chain — only ScrollbarOrientation and thumb\_symbol vary. Two copies of a widget-config dance that ratatui already parameterizes.
  _Fix:_ Collapse into one \`fn draw\_scrollbar(frame, track, horiz: bool, total, pos, viewport)\` that picks orientation + thumb symbol from the bool; the two callers pass \`false\`/\`true\`. Keeps the thumb-glyph constants as the single source of truth.
- ⚪ **[LOW]** \[maintainability\] tree\_bars and content\_bars duplicate the two-pass needs\_v/needs\_h computation — `src/presenter.rs:278` · 1/4 models _(maintainability)_
  Both helpers carry the identical v0/h0/needs\_v/needs\_h four-line two-pass dance (vbar steals width → maybe hbar, hbar steals height → maybe vbar); only the \`max\_w\` input (gated by wrap in content\_bars) differs. The two-pass logic is the subtle part worth having exactly once.
  _Fix:_ Extract \`fn needed\_bars(area: Rect, total: usize, max\_w: usize) -\> (bool, bool)\` returning (needs\_v, needs\_h); each helper computes its \`max\_w\` then calls \`bar\_layout(area, …needed\_bars(…))\`, removing the duplicated overflow reasoning.
- ⚪ **[LOW]** \[correctness\] wrapped prose can overflow vertically without a scrollbar — `src/presenter.rs:312` · 1/4 models _(correctness)_
  \`content\_bars\` decides whether to draw a vertical scrollbar from the raw line count (\`total\_lines\`), but under \`wrap\` the actual scrollable extent is the wrapped row count. A long wrapped markdown file with few raw lines can therefore overflow vertically with no scrollbar, contradicting the documented "scrollbar whenever there is more to see than fits".
  _Fix:_ Decide the content vertical scrollbar from the wrapped row count (e.g. \`rendered\_line\_count()\`) when \`wrap\` is true, so the bar reflects the real wrapped extent; keep the cheap raw-line path for the common unwrapped case.

- **[MEDIUM]** \[correctness\] Dragging the tree's vertical scrollbar does not track the pointer — the thumb stays pinned then jumps — `src/controller.rs:1016`
  _Dismissed:_ FIXED at a392d4a: draw\_tree now drives the tree vertical thumb by state.selected (the cursor), matching scroll\_tree\_to\_row's cursor mapping, so the thumb tracks the drag. Confirmed by code-read and the new regression test tree\_vertical\_thumb\_tracks\_the\_selection (thumb moves down as the selection moves down).
- **[MEDIUM]** \[correctness\] Wrapped content can scroll without a scrollbar — `src/presenter.rs:312`
  _Dismissed:_ FIXED at a392d4a: the content vertical scrollbar is now sized/positioned from ViewState.content\_rows = controller rendered\_line\_count\_for(wrap) (wrapped-aware), not raw content.lines.len(). One source of content height now. Confirmed by code-read + test content\_vertical\_scrollbar\_is\_driven\_by\_rendered\_rows\_not\_raw\_lines.
- **[MEDIUM]** \[correctness\] Content vertical scrollbar is hidden under wrap when wrapped rows overflow but raw lines fit — `src/presenter.rs:410`
  _Dismissed:_ FIXED at a392d4a (same content\_rows change): a single long line that wraps past the viewport now shows a vertical bar; the thumb sizes/positions against wrapped rows. Confirmed by the regression test (one raw line, content\_rows=60 -\> bar present; content\_rows=5 -\> none).

smarzban:
## Orchestrator review — re-gate verdict: ✅ Approve

This is the **re-gate** after the PR grew well past its first pass (it now carries the in-pane
draggable scrollbars, tree h-scroll, four drag targets, and the inside-bar/thumb-only layout).
Round 1 ran the full panel — deterministic scan + **holistic ×4** (kimi · glm · opus · gpt-5.5) +
**lens-simplify ×2** (gpt-5.5 · glm). It found three real correctness **mediums**, all now fixed in
`a392d4a`:

- **Content vertical scrollbar wrong/missing under wrap** (4 models). It sized/positioned from the
  raw line count, so a wrapped file (few long lines) pinned the thumb early or showed no bar. Fixed:
  the bar is now driven by the controller's wrapped-aware rendered-row count (`content_rows`).
  Regression test: `content_vertical_scrollbar_is_driven_by_rendered_rows_not_raw_lines`.
- **Tree vertical thumb didn't track the drag** (2 models). The thumb followed the viewport offset
  while the drag scrubs the selection, so it stayed pinned then jumped. Fixed: the tree thumb is now
  driven by the cursor, matching the drag. Regression test: `tree_vertical_thumb_tracks_the_selection`.
- Plus the low-severity correctness nits: a scrollbar drag now clears the pending double-click, and
  a press always (re)sets the drag target so a missed release can't leave a stale drag. The tree-vbar
  scrub also skips the (expensive) re-render when the row is unchanged (perf nit, opus + glm).

**Deferred (advisory, lens-simplify low/maintainability):** four DRY cleanups — merge
`draw_v/hscrollbar`, dedup the two-pass overflow, dedup the `scroll_*_to_*` helpers + the Down/Drag
match. Correct as-is and unit-tested; refactoring under the gate adds risk for no behavior change.
Logged as follow-up.

**On the verdict's evidence — disclosed plainly:** the round-2 model verifiers (codex, then kimi,
then opus on a terse prompt) all returned **unparseable non-votes** — three different backends failing
identically points to an environment/tooling issue (consistent with glm's `ANTHROPIC_API_KEY`
connector error earlier this run), **not** a clean pass. So I did **not** rest the verdict on a
non-vote. Resolution was confirmed by (a) a line-by-line orchestrator code-read of all six fixes,
(b) **two new regression tests** that directly assert the two medium fixes, and (c) the full suite
green (**350 tests**, `clippy -D warnings` + `fmt` clean). That is the objective basis for the pass.

**Decision: ✅ Approve.** Not merging — yours to merge after you're satisfied.

PR #51, opened 2026-06-26
feat: go to file (fuzzy find) — `f` quick-open finder

smarzban:
## Go to file (fuzzy find)

Adds a keyboard-summoned **fuzzy file finder** to the viewer. Today the only way to reach a file is
to scroll and expand the tree by hand — slow in a large/deep repo, and it assumes you already know
where the file lives. Press **`f`** to open a modal overlay, type part of a filename, and jump
straight to it (the quick-open / Ctrl-P interaction from fzf, `fd`, and editors).

As you type, a **hand-rolled fuzzy matcher** filters a **flat, gitignore-aware file index** of the
project; **↑/↓** move the selection, **Enter** jumps (reveals the file in the tree — ancestors
expanded, cursor set, scrolled into view — and renders it via the normal view policy), **Esc**
cancels. `q` and other printable keys are literal query text. Fully **read-only**: jumping is pure
navigation — no file or git mutations, no persistent state.

Tracking: Linear **SMA-50** · project *Go to file (fuzzy find)*. Spec chain (Brief → Acceptance
Criteria → Design → Tech Stack → Plan) drove this PR; built via the agent-sdlc pipeline.

### Highlights
- **Zero new Cargo dependencies** — the fuzzy matcher is hand-rolled (case-insensitive subsequence,
  **basename-weighted** so `apprs` surfaces `src/app.rs` above a deep `…/apple_response.rs`); the
  overlay reuses the existing worktree-picker shell (centered modal, bordered list, in-pane
  scrollbar, control-byte sanitization, modal input gating).
- **Flat File Index** ([ADR-0005](docs/adr/0005-file-index.md)) — a recursive, gitignore-aware
  `ignore`-crate walk, **independent of the Tree Model's** lazy expansion and view filters (so the
  finder sees the full project set regardless of the `c`/`.` tree toggles). Always skips `.git/`.
  Built fresh on each open (no cache, reflects current disk state). Scope is a single knob defaulting
  to gitignore-respecting, so a future Settings feature (SMA-49) can widen it without rework.
- **Reusable Prompt-Input primitive** (`src/prompt.rs`) — the query buffer, shared with the
  forthcoming in-file-navigation feature rather than inlined.
- **Render on Enter only** (no live preview); empty query shows a "type to find a file" placeholder.

### Post-build refinements (owner live-testing on the real herdr host)
- **Zoom on confirm when only the tree is visible** — in a narrow split pane the layout is tree-only;
  a confirmed jump now opens the file in zoom mode (mirrors the tree's Enter/activate) instead of
  leaving it revealed off-screen.
- **Mouse support inside the finder overlay** — wheel scrolls the selection, click selects,
  double-click confirms; the worktree picker stays inert behind it.
- **Horizontal scroll** (←/→ + horizontal wheel) for long result paths in a small pane, and the
  duplicate top `esc cancel` chip dropped (footer is the single home for key hints).

### Acceptance criteria (28)
All 28 ACs are covered by tests except **AC-N4** and **AC-N6** (reviewer-checked, Spec Conformance):

**Opening** — AC-1 (`f` opens the modal), AC-2 (empty query → placeholder, no rows).
**Matching** — AC-3 (case-insensitive subsequence), AC-4 (basename-weighted ranking), AC-5 (rows show
root-relative path), AC-6 (no match → no rows; Enter is a no-op, finder stays open).
**Keyboard** — AC-7 (printable keys edit the query, not viewer actions — e2e), AC-8 (↑/↓ move
selection), AC-9 (Esc cancels, nothing changes; `q` is literal).
**Jump** — AC-10 (Enter reveals in tree: expand ancestors + set cursor + scroll), AC-11 (renders via
the normal view policy).
**Index & scope** — AC-12 (all files under root, recursive), AC-13 (ignored files absent),
AC-14 (`.git` never a candidate), AC-15 (files only, no dirs), AC-16/AC-17 (independent of the `c`
and `.` tree filters), AC-18 (rebuilt each open).
**Degradation** — AC-19 (works with or without git), AC-20 (missing-at-confirm → non-fatal notice, no
crash, no partial jump).
**Performance** — AC-21 (ready to accept input within 1s on 10k files), AC-22 (each keystroke
re-filters within 300ms; matching never blocks input).
**Negative** — AC-N1 (no fs mutation), AC-N2 (no git mutation), AC-N3 (path/name only, never
contents), AC-N4 (no persistent state — *reviewer-checked*), AC-N5 (never above the root),
AC-N6 (opens only via explicit `f`, no event hook — *reviewer-checked*).

### Task → criterion coverage (12 tasks, test-first)
| Task | Module(s) | Advances |
| --- | --- | --- |
| T-1 Fuzzy Matcher | `src/fuzzy.rs` | AC-3, AC-4, AC-6, AC-N3 |
| T-2 File Index | `src/index.rs` | AC-12,13,14,15,18,19,N1,N2,N5 |
| T-3 Prompt-Input | `src/prompt.rs` | AC-7 |
| T-4 Tree `reveal(path)` | `src/tree.rs` | AC-10, AC-20, AC-N5 |
| T-5 OpenFinder intent + `f` + open | `src/intent.rs`,`src/input.rs`,`src/finder.rs`,`src/controller.rs` | AC-1, AC-18, AC-N1, AC-N3, AC-N6 |
| T-6 Keystroke match + ↑/↓ nav | `src/finder.rs`,`src/controller.rs` | AC-6, AC-7, AC-8 |
| T-7 Confirm / cancel / no-op | `src/controller.rs`,`src/tree.rs` | AC-6, AC-9, AC-10, AC-11, AC-20 |
| T-8 Finder overlay | `src/presenter.rs` | AC-1, AC-2, AC-5 |
| T-9 Event-loop raw-key routing + AC-7 e2e | `src/app.rs` | AC-7 |
| T-10 Scope independence + non-git | tests | AC-16, AC-17, AC-19 |
| T-11 Performance | tests | AC-21, AC-22 |
| T-12 Negative criteria & conformance | tests | AC-N1,N2,N4,N5,N6 |

### Tests & green bar
**441 tests** pass (0 failures) — baseline 352 + 89 new across `tests/{fuzzy,index,index_perf,tree,controller,presenter,presenter_narrow,e2e_keyboard}.rs` and 2 insta snapshots. `cargo clippy
--all-targets -- -D warnings`, `cargo fmt --check`, and `cargo build --release` all clean.

### Scope / constraints
- Read-only constitution upheld; ephemeral state only (no persistent store).
- Touches only `src/` + `tests/`. No herdr manifest/event surface added (AC-N6).
- 16 commits: T-1…T-12 (one atomic green commit per task) + 4 reviewed post-build commits.

### Not in this PR
- **Real-mouse forwarding from herdr into the pane** — a separate herdr-integration concern (the
  viewer requests mouse capture and handles click/scroll correctly; the gap is upstream
  forwarding). Tracked separately, not bundled here.
- **Settings UI** to widen the search scope (SMA-49) and **live preview** as the selection moves are
  explicit non-goals / deferred enhancements.


smarzban:
## Review Gate — Round 1

Findings: 11 total — 0 critical · 0 high · 3 medium · 7 low · 1 info.
_Reviewed by:_ holistic + lens-security + lens-spec + tools · models: deterministic, glm-5.2:cloud, claude-opus-4-8, gpt-5.5

### Advisory (non-blocking)
- ⚪ **[LOW]** \[correctness\] Finder keystroke/nav between two same-row clicks is misread as a double-click confirm — `src/controller.rs:1712` · 1/3 models _(correctness)_
  handle\_finder\_click stores last\_click on a single click, and is\_double\_click matches on the same screen ROW within 400ms (column-agnostic). The keystroke/nav arms of handle\_finder\_key (Char, Backspace, Up/Down, Left/Right) never clear last\_click, so the sequence click-row → type-or-arrow → click-same-row (all within 400ms) is treated as a double-click and confirms/opens a file the user only single-clicked — often a \*different\* file than the first click, since typing changed the match list. This is the exact cross-contamination class the suite guards for the open/Esc/confirm paths (scenario\_a/b), but the keystroke/nav vector is unguarded.
  _Fix:_ Clear self.last\_click = None in the Char/Backspace/Up/Down/Left/Right arms of handle\_finder\_key (or, more robustly, in FinderState::push/backspace/move\_selection callers), mirroring the Esc/confirm/open clears — any query or selection change should reset the pending double-click.
- ⚪ **[LOW]** \[maintainability\] re\_root closes the picker but not the finder (asymmetric teardown relies on a fragile invariant) — `src/controller.rs:447` · 1/3 models _(maintainability)_
  \`re\_root\` resets \`self.picker = None\` (AC-13) but leaves \`self.finder\` untouched. Today this is safe only because re\_root is reachable solely via the picker-confirm path (controller.rs:833), and the picker can never open while the finder is open (modal exclusion in \`handle()\`). The finder field doc does not document this invariant, so a future change that makes re\_root fire while the finder is open (a focus-gained auto-switch, an external trigger, or routing re-root through another intent) would leave a \`FinderState\` holding candidates from the OLD root; \`confirm\_finder\` would then \`self.root.join(old\_rel)\` against the new root, producing a non-existent path that \`reveal\` rejects with a confusing "Could not open" notice rather than a clean close.
  _Fix:_ Add \`self.finder = None;\` next to \`self.picker = None;\` in \`re\_root\` (and clear \`last\_click\` for symmetry), and note in the finder field doc that re\_root closes it. This makes the modal-closed invariant structural rather than incidental.
- ⚪ **[LOW]** \[maintainability\] handle() enforces picker modality but not finder modality (asymmetric guard) — `src/controller.rs:741` · 1/3 models _(maintainability)_
  handle() short-circuits to handle\_picker\_intent while self.picker.is\_some(), making the picker modal at the controller layer, but there is no equivalent self.finder.is\_some() guard. Finder modality rests entirely on the app.rs event-loop routing (the controller.finder\_open() arm). Any future or test caller that invokes handle() while the finder is open would leak intents to the tree beneath it and could open a second modal (OpenFinder/SwitchWorktree), since draw/geometry happily render both overlays at once.
  _Fix:_ Add a symmetric guard at the top of handle() (e.g. \`if self.finder.is\_some() { return Effects::noop(); }\` or route to a finder handler), so the finder's modality is enforced at the same layer as the picker's rather than only in the run loop.
- ⚪ **[LOW]** \[correctness\] Non-ASCII case-insensitive matches are not supported — `src/fuzzy.rs:27` · 1/3 models _(correctness)_
  AC-3 asks for case-insensitive subsequence matching, but the matcher uses \`to\_ascii\_lowercase()\` for both query and path. Paths containing non-ASCII letters such as \`Résumé.md\` will not match queries with different case such as \`résumé\`.
  _Fix:_ Use Unicode-aware case folding for query and candidate characters, or explicitly constrain the spec/tests to ASCII-only matching.
- ⚪ **[LOW]** \[test-coverage\] AC-18 reopen freshness is only partially covered — `tests/controller.rs:3122` · 1/3 models _(test-coverage)_
  AC-18 requires the finder index to be rebuilt each time the finder opens, including files created or removed since the previous session. The tests cover \`index::build\` seeing a newly-created file, but not a same-controller finder close/reopen flow or the removed-file absence half.
  _Fix:_ Add a controller test that opens and closes the finder, creates one file and removes another, reopens it, then asserts the new candidate is present and the removed candidate is absent.
- ⚪ **[LOW]** \[test-coverage\] AC-20 lacks unreadable-file coverage — `tests/controller.rs:3244` · 1/3 models _(test-coverage)_
  The AC-20 test covers a candidate removed after opening the finder, but not the stated unreadable-file case. A test would currently need to assert that an unreadable matched file leaves the selection unchanged and surfaces a notice.
  _Fix:_ Add a platform-gated test that opens the finder, makes the matched file unreadable before Enter, and asserts no selection change plus a non-fatal notice.
- ⚪ **[LOW]** \[test-coverage\] AC-N3 does not prove content is ignored — `tests/fuzzy.rs:65` · 1/3 models _(test-coverage)_
  AC-N3 specifically says a query found only inside file contents must yield no match, but the current test only checks a string that appears in neither path nor content. A content-scanning regression could satisfy that test while violating the criterion.
  _Fix:_ Add an integration/controller test with a file whose contents contain a unique query string that is not a subsequence of any path, then assert the finder has zero matches.
- ⚪ **[INFO]** \[performance\] Finder index build runs synchronously on the input thread, unlike rendering — `src/controller.rs:1651` · 1/3 models _(performance)_
  open\_finder calls index::build (a full recursive gitignore-aware filesystem walk, no depth limit) on the input thread, and every keystroke runs fuzzy::match\_and\_rank synchronously too. AC-21/AC-22 bound these to \~1s/300ms at 10k files (and perf tests guard that), but the codebase deliberately moved the comparably-heavy render/diff work off-thread (AC-23). On a monorepo well beyond 10k files, pressing \`f\` (and each keystroke) freezes the UI — an architectural inconsistency with the off-thread render precedent, even though it is within the current accepted scope.
  _Fix:_ Acceptable for now given the AC budget; if larger repos are in scope later, dispatch index::build (and optionally match\_and\_rank) to a worker the way dispatch\_render does, surfacing results via poll().

- **[MEDIUM]** \[security\] unbounded finder indexing can exhaust resources on hostile trees — `src/index.rs:33`
  _Dismissed:_ Three codex findings on the read-only index walk, all single-model, code-checked as non-blocking for a LOCAL read-only file viewer. (a) Unbounded-indexing DoS: the 'attacker' is the local repo the USER chose to open in a TUI, not remote/untrusted input; AC-21/AC-22 scope the perf budget to 10k files and Brief decision \#6 deliberately ships v1 with no result cap ('revisit only if a huge repo feels slow'); the \`ignore\` walk is OS/filesystem-bounded. (b) Symlinks omitted: DELIBERATE — index.rs filters \`is\_file()\` and WalkBuilder's default \`follow\_links(false)\` means a symlink's entry type is symlink (not file), so it is excluded BY DESIGN as the AC-N5 escape guard (a followed symlink could resolve above root). Intended behavior, not a defect. (c) Non-UTF-8 filenames: \`to\_string\_lossy\` (index.rs:41) renders invalid-UTF-8 names with U+FFFD; confirm's \`root.join(rel)\` then misses the real path so \`reveal\` returns false → AC-20's graceful path (render.rs:57/80 non-fatal placeholder, no crash, no mutation, recoverable); rare in a browsed repo; a proper fix is an OsString-identity refactor disproportionate to the rarity. Tracked as follow-ups; none is a merge-blocker.
- **[MEDIUM]** \[performance\] Finder overlay rebuilds the full match list three times per frame, unbounded by the visible window — `src/presenter.rs:956`
  _Dismissed:_ Single-model (glm), contested. Real inefficiency confirmed in code: finder\_overlay\_layout builds a full Vec\<Line\> over ALL matches (presenter.rs:956) to size-to-content, draw\_finder\_overlay rebuilds it (presenter.rs:1067), and finder\_view clones all matched paths (controller.rs:708) — \~3-4x full-list materialization per frame; only the final \`window\` is bounded by skip/take. NOT an AC-22 violation: at the 10k scope \~40k short (\~40-char) sanitize\_label String allocations is low-single-digit ms, far under AC-22's 300ms input-responsiveness budget which the (perf-tested) fuzzy scan dominates; 60fps rendering is not an AC. Real optimization (reduce the redundant rebuilds), tracked as a perf follow-up with the off-thread-index note; not a merge-blocker.
- **[MEDIUM]** \[correctness\] Unreadable finder targets still change selection — `src/tree.rs:311`
  _Dismissed:_ Single-model (codex lens-spec), contested. Real but PARTIAL AC-20 gap: reveal() guards 'missing' via is\_file() (tree.rs:311), but an existing-but-unreadable file (chmod 000) passes is\_file(), so confirm moves the selection + dispatches render. HOWEVER the 'non-fatal, no crash' half of AC-20 IS met: render.rs:57/80 returns a Binary placeholder for an unreadable file ('unreadable (e.g. permissions) -\> placeholder, not a misleading empty pane') — no crash, no mutation (read-only), recoverable. The only strict deviation is 'stays on current selection' for the unreadable sub-case, which would need a racy pre-read permission probe in reveal disproportionate to a rare chmod-000 file in a user-browsed repo. Conformance nuance, tracked as a follow-up (with the LS3 coverage gap); not a merge-blocker.

smarzban:

**Panel:** deterministic scan (clean) + holistic ×3 (glm-5.2, claude-opus-4-8, gpt-5.5) + lens-spec (gpt-5.5) + lens-security (gpt-5.5). I read the code behind every gating finding myself before adjudicating.

**Coverage caveat (flagged, not hidden):** 3 reviewer passes failed in this environment and their coverage is lost — `kimi-k2.7-code` (holistic) and **both opus lens passes** (lens-spec exited 1, lens-security returned an unparseable non-vote). The ollama `kimi` failure is the known `ollama launch claude` connector issue; the opus lenses failed transiently on the longer prompts. Holistic still ran on 3 models (not thin), with cross-lineage diversity (open: glm; closed: opus + codex), and the lenses ran on codex — but the second lens shot was not obtained, so the lens findings below are single-model.

### What the PR implements
An `f`-summoned fuzzy file finder: a modal overlay with a query line + ranked result list over a flat, gitignore-aware **File Index** (separate from the Tree Model — ADR-0005), a hand-rolled basename-weighted subsequence matcher (zero new deps), reveal-in-tree + render-on-Enter, Esc-cancel, and mouse support (wheel/click/double-click). Read-only throughout. 441 tests; clippy `-D warnings`, fmt, release all clean.

### Verdict: ✅ PASS (gating) — 0 blocking
No critical/high findings, no tool findings, and no medium corroborated by ≥2 models. The three medium clusters are each **single-model** and I dismissed them with code-checked justifications (see the gate comment above): the index-walk bundle (unbounded-indexing "DoS" → local user-chosen repo + spec's deliberate no-cap at the 10k scope; symlinks-omitted → **deliberate** AC-N5 escape guard; non-UTF-8 names → lands in AC-20's graceful no-crash placeholder path), the overlay 3–4× re-materialization (real inefficiency, but ~low-ms at 10k, far under AC-22's 300ms which the perf-tested fuzzy scan dominates), and the AC-20 "unreadable-but-existing file moves selection" partial-conformance (no crash — `render.rs` shows a Binary placeholder; only the strict "stay-put" sub-case is unmet, and that would need a racy permission probe).

### Real advisory findings worth addressing (surfaced for owner sign-off — NOT silently accepted)
These don't block the gate (low severity / single-model / recoverable in a read-only viewer), but two are genuine and cheap, and I recommend folding them in **before merge**:

1. **[correctness] `f`-finder double-click misfire** (`controller.rs` `handle_finder_key`) — a *click → type/arrow → click-same-row* within 400 ms is misread as a double-click and opens a file the user only single-clicked (often a *different* file, since typing changed the list). Same cross-contamination class the suite already guards on the open/Esc/confirm paths; the keystroke/nav arms just don't clear `last_click`. Fix is ~one line per arm + a regression test. *Recoverable (read-only) but a real UX glitch you'll likely hit in live testing.*
2. **[robustness] finder modality is only structural in `app.rs`, not the controller** (corroborated by opus + glm) — `handle()` guards the picker but not the finder, and `re_root` clears the picker but not the finder. Safe-by-construction today (mutual exclusion holds), but a trivial defensive fix (`if self.finder.is_some()` guard in `handle()` + `self.finder = None` in `re_root`) makes the invariant structural.
3. **[test-coverage] two cheap, clean test adds** — AC-N3 "a content-only string yields no match" (proves the negative criterion directly) and AC-18 "same-controller close/reopen sees created + dropped files."

### Tracked as follow-ups (real, but non-trivial or deliberately out of v1 scope)
- AC-20 **unreadable-but-existing** file: tighten to leave the selection unchanged (currently graceful via the content placeholder).
- **Non-UTF-8 filenames** openable via the finder (needs an `OsString`-identity refactor, not a lossy display string).
- **Finder perf at scale beyond 10k**: reduce the 3–4× per-frame match-list re-materialization, optionally move `index::build` off-thread (mirrors the render worker), and consider a candidate cap/truncation notice for pathological trees — one theme.
- **Unicode case-folding** in the matcher (today ASCII-only; AC-3 didn't mandate Unicode).

### Dismissed as intended behavior
- Symlinks excluded from the index — deliberate AC-N5 escape guard, not a defect.

**Decision: ✅ Approve from a gating standpoint.** Merge and release remain **owner-gated** — ship does not merge or tag. I have **not** pushed any fix to this branch (it's your outward artifact). Recommend a quick optional hardening pass for items 1–3 above before you merge; the rest are tracked follow-ups.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ tools + verify · models: deterministic, gpt-5.5

### Progress since Round 1
✅ Resolved (0)
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:
## Orchestrator review — Round 2 (verification) · ✅ Approve

Round 2 is single-model **verification** of the round-1 hardening, not a fresh discovery panel. After the owner green-lit the optional hardening, I implemented the round-1 advisory fixes in `448a230` and verified them.

**What changed in `448a230`** (touches only `src/controller.rs` + `tests/controller.rs`):
- **O1 — finder double-click reset:** `handle_finder_key` now clears `last_click` after a keystroke/nav (Char/Backspace/↑↓/←→), so a finder click → type → click-same-row within 400 ms is no longer misread as a double-click confirm.
- **O2 — finder modality:** `handle()` now returns `noop` while the finder is open (symmetric with the picker guard) — no intent can leak to the tree or open a second modal beneath the overlay.
- **G2 — `re_root` teardown:** closes the finder + clears `last_click` alongside the picker (defensive; unreachable today).
- **Coverage:** +O1 keystroke double-click regression, +O2 intents-inert, +AC-N3 content-blindness (end-to-end index→matcher), +AC-18 same-controller reopen (created + dropped files). Two existing tests that relied on intents leaking through an open finder were adjusted (close the finder before the `handle()`-driven intent) — preserving their original assertion intent.

**Verification:** deterministic scan clean, and a single-model verifier (codex `gpt-5.5`) confirmed **all four items resolved with no regression** (`[]`). The opus verifier returned a non-vote (unparseable) — per the non-vote guard I did **not** treat that as a pass; codex provided the authoritative verification. Locally: **445 tests** (+4), clippy `-D warnings`, fmt, release all clean.

### Cumulative summary
**Implements:** an `f`-summoned fuzzy file finder — flat gitignore-aware index (ADR-0005), hand-rolled basename-weighted matcher (zero new deps), reveal-in-tree + render-on-Enter, mouse support, read-only throughout. 28 ACs.

**Fixed across the gate:** O1 (double-click misfire), O2 + G2 (finder modal symmetry), and two test-coverage gaps (AC-N3 content-blindness, AC-18 reopen).

**Deferred / not covered (tracked follow-ups, none blocking):** AC-20 unreadable-but-existing file (graceful placeholder today; strict stay-put would need a racy permission probe); non-UTF-8 filenames openable (OsString-identity refactor); finder perf beyond 10k (off-thread index + reduce the 3–4× per-frame match-list rebuild + optional candidate cap); Unicode case-folding. **Coverage caveat from round 1:** the `kimi` holistic pass and both opus lens passes failed in this environment (connector / non-vote) — holistic still ran on 3 models with cross-lineage diversity, lenses on codex.

**Decision: ✅ Approve.** Both rounds PASS; the round-1 advisory findings are fixed and verified. **Merge and release remain owner-gated** — the gate does not merge or tag. PR #51 is hardened and ready for your testing + merge.

PR #52, opened 2026-06-27
feat: tree borders show dir name + branch (SMA-249); fix picker hscroll overshoot (SMA-229)

smarzban:
## Tree borders show context + picker hscroll fix

Two small UI fixes from live testing / the backlog, bundled into one PR (the changes interleave in `presenter.rs`/`controller.rs`).

### SMA-249 — tree pane borders reflect context (not a static "Files")
The tree column's **top border** is now the **root directory basename** (e.g. `herdr-plugin`) instead of the static "Files" — mirroring how the content pane already titles itself from the selected node, and auto-updating after a `W` worktree switch. The **bottom border** shows the **current git branch** when in a repo, and is omitted entirely outside a repo / on a detached HEAD (degrade gracefully).

- `ViewState` gains `root_name` + `branch`. The branch is **cached on the controller** (computed once per (re-)root from the resolved repo root — never per-frame, since it only changes on a re-root); `git::current_branch` is now `pub`.
- The title is sanitized (AC-27 — dir/branch names are untrusted) and truncated with an ellipsis so a long name can't break the border.

### SMA-229 — worktree picker horizontal scroll no longer overshoots
The picker's `→`/`←` scroll had the same latent bug the finder did: `→` (`Intent::Expand`) incremented `hscroll` monotonically with the clamp only at draw, so over-scrolling right parked the stored offset past the real max and `←` needed several presses before the view moved. Fixed by **mirroring the finder fix**: extract `picker_overlay_layout` (shared by `draw_picker_overlay` + `geometry`) exposing `max_hscroll`, feed it back as `PaneGeometry.picker_max_hscroll`, and clamp the stored `PickerState.hscroll` in `set_pane_geometry` each frame. (The picker is keyboard-only / mouse-inert, so no scrollbar-drag parity is needed — unlike the finder.)

### Tests & green bar
**453 tests** pass (0 failures); `clippy --all-targets -- -D warnings`, `cargo fmt --check`, `cargo build --release` all clean.
- New: tree title/branch render (root name shown, branch shown vs omitted, long-name truncation), controller `root_name`/`branch`, picker hscroll overshoot regression.
- 5 tree snapshots re-accepted — title-only change ("Files" → root name), verified line-by-line.
- Touches **only** `src/` + `tests/`.

### Provenance / review note
The implementation was drafted by a background subagent that **exited mid-task** (its process died). I reviewed the **entire** resulting diff myself (it had landed a compiling, correct core), then finished the gaps — the `tests/controller.rs` `PaneGeometry` compile fix, one missed `"Files"` test assertion the subagent didn't update, the four new tests, and the five snapshot re-accepts — and verified the full green bar. So this is conductor-reviewed, not unreviewed agent output.

### Not in this PR
Owner-gated: please live-test before merge. Deferred backlog (unchanged): SMA-230 (AC-20 unreadable), SMA-231 (non-UTF-8), SMA-232 (finder perf >10k), SMA-233 (Unicode folding), and the rest of the tech-debt (SMA-59 the one quick security win).


smarzban:
## Review Gate — Round 1

Findings: 3 total — 0 critical · 0 high · 1 medium · 2 low · 0 info.
_Reviewed by:_ holistic + lens-security + tools · models: deterministic, kimi-k2.7-code:cloud, glm-5.2:cloud, claude-opus-4-8, gpt-5.5
⚠️ **Coverage:** 6/8 planned reviewer passes voted — lost: lens-spec/(skipped) (no formal AC spec — SMA-249/229 are prose backlog issues; conformance checked via the holistic panel + owner live-test, not an independent spec lens); scan:secrets/gitleaks (gitleaks not on PATH — secret scan skipped (diff is Rust src/tests only, low secret risk))

### Must fix
- 🟠 **[MEDIUM]** \[correctness\] Branch label stays stale after git refresh — `src/controller.rs:1960` · 3/4 models _(correctness)_
  \`refresh\_git\_state()\` is used on focus regain, editor return, and explicit refresh to pick up external git changes, but the new \`current\_branch\` cache is only populated at construction and re-root. If the user checks out a different branch in the same worktree while the viewer is open, the tree can show the old branch while the rest of the git-derived state has been refreshed.
  _Fix:_ Store the resolved repo root or expose \`current\_branch\` through the \`GitService\`, then update \`self.current\_branch\` inside \`refresh\_git\_state()\` alongside status/changed-set refreshes; add a test that switches branches and calls \`handle\_focus\_gained()\` or refresh.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[security\] Unicode format characters survive label sanitization — `src/presenter.rs:155` · 1/4 models _(security)_
  \`sanitize\_label\` only drops \`char::is\_control()\`, so repo-controlled branch or directory names can include bidi controls and zero-width format characters such as U+202E/U+2066 or U+200B. ratatui filters C0/C1 controls but preserves these zero-width graphemes into the buffer/backend, allowing a crafted border title to visually reorder or hide text in the terminal.
  _Fix:_ Strip or visibly escape Unicode format/default-ignorable characters used for bidi and zero-width behavior before rendering labels, and add tests that U+202A..U+202E, U+2066..U+2069, U+200B/U+200C/U+200D, and BOM do not survive in rendered buffer output.
- ⚪ **[LOW]** \[maintainability\] truncate helpers measure char count, not display columns — `src/presenter.rs:181` · 3/4 models _(maintainability)_
  truncate\_title/truncate\_middle use chars.len() as the width budget, but the rest of the layout (Line::width) counts unicode display columns. A wide-char (CJK) or combining-mark root/branch name can still exceed the column budget; the title won't break the border (ratatui clips it) but the ellipsis can land off-screen, so the intended 'fits with a … indicator' guarantee doesn't hold for those names. Char-index slicing itself is boundary-safe — no panic.
  _Fix:_ Budget against display width (e.g. unicode-width / sum of Line widths) rather than chars.len(), or document that the helpers are an approximation and rely on ratatui clipping for the wide-char tail (matching draw\_content's untruncated title).

smarzban:

**Panel:** deterministic scan (clean) + holistic ×4 (kimi, glm, opus, codex — all 4 voted) + lens-security (codex). I read the code behind the gating finding. **Coverage:** lens-spec was not run (SMA-249/229 are prose backlog issues, no formal AC chain — conformance is covered by the holistic panel + your live-test, just not an independent spec lens), and gitleaks isn't installed (secret scan skipped; the diff is Rust src/tests only). Both are flagged in the gate's Coverage line.

### The one real blocker — fixing it
**Stale branch label after an external checkout + refresh** (`controller.rs` `refresh_git_state`). Confirmed in the code: `refresh_git_state` re-queries status + the changed-set but **never updates `self.current_branch`**, which I only cache at construction and on re-root. So `r` / editor-return / focus-gain after a `git checkout` in the same worktree would leave the tree's bottom border showing the old branch. Three models flagged it (codex medium, glm/opus low) — a genuine staleness bug the cache introduced. **Decision: 🔴 Request changes.** Fix: recompute `current_branch` inside `refresh_git_state` (a single `git rev-parse` from the root, which git resolves to the repo) + a regression test that switches branches and refreshes.

### Advisory (non-blocking) — deferring, with code-checked reasons
- **Truncation budgets by char count, not display width** (low ×3). Real but minor: the reviewer itself confirmed the char-index slicing is **panic-safe** — the only effect is that for a wide-char (CJK) name the ellipsis can sit a column off. Repo/branch names are overwhelmingly ASCII (1 col/char), and a display-width fix would need the `unicode-width` crate — a **new dependency this project deliberately avoids**. Deferring (document/approximation); not worth a dep for a cosmetic wide-char edge.
- **`sanitize_label` lets Unicode bidi/zero-width *format* chars through** (low ×1, lens-security). Real but **pre-existing** — `sanitize_label` strips `is_control()` only and is used for *every* label (tree rows, picker, finder, content); this PR merely routes one more name (the tree title) through it. Impact is visual spoofing of a *title* on a local read-only viewer with no egress (no clipboard/subprocess in this path). Deferring as a **separate, broad `sanitize_label` hardening follow-up** (strip default-ignorable/bidi chars across all labels), not a rider on this PR.

Implementing the branch-refresh fix now, then round 2 verifies it + checks for regressions.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ tools + verify · models: deterministic, gpt-5.5
⚠️ **Coverage:** 2/4 planned reviewer passes voted — lost: verify/claude-opus-4-8 (unparseable non-vote — fell back to codex gpt-5.5 (the reliable parser), which verified the fix); scan:secrets/gitleaks (gitleaks not on PATH — secret scan skipped (Rust src/tests only))

### Progress since Round 1
✅ Resolved (1): \[correctness\] Branch label stays stale after git refresh — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:
## Orchestrator review — Round 2 (verification) · ✅ Approve

Round 2 is single-model **verification** of the round-1 fix, not a fresh sweep. Deterministic scan clean; the verifier confirmed the fix and found no regression.

**What was fixed (cumulative):** Round 1's panel (4/4 holistic + lens-security) caught one real blocker — the **stale branch label**: `refresh_git_state` re-queried status but never updated the cached `current_branch`, so an external `git checkout` + refresh left the tree's bottom border showing the old branch. Fixed in `3521662` by recomputing `current_branch` (a single `git rev-parse` from the root) inside `refresh_git_state`, with a regression test that checks out a new branch and refreshes.

**Verification:** the codex verifier returned `[]` — blocker resolved at HEAD, no regression. The opus verifier returned an unparseable non-vote, so per the non-vote guard I did **not** count it as a pass and fell back to codex (the reliable parser); the lost opus shot is flagged in the Coverage line. Locally: **455 tests** pass, clippy `-D` / fmt / release clean.

### Cumulative summary
**Implements:** (SMA-249) tree top border = root dir basename, bottom border = current branch with middle-ellipsis truncation (both ends visible when narrow), sanitized, omitted outside a repo / cached per-(re-)root **and now per-refresh**; (SMA-229) worktree picker hscroll clamped in state via `picker_max_hscroll` geometry feedback (mirrors the finder fix).

**Fixed across the gate:** the stale-branch-after-refresh bug (R1 → fixed R1-commit, verified R2).

**Deferred / not covered (advisory, non-blocking — tracked, not in this PR):**
- `truncate_*` budget by char count, not display width (low ×3) — a display-width fix needs the `unicode-width` crate, a new dep this project avoids; the char-slicing is panic-safe (reviewer-confirmed), so the only effect is a cosmetic ellipsis offset for CJK names.
- `sanitize_label` passes Unicode bidi/zero-width *format* chars (low ×1, lens-security) — **pre-existing** and repo-wide (all labels), not introduced here; impact is title visual-spoofing on a local read-only viewer with no egress. Worth a separate broad `sanitize_label` hardening follow-up.

**Coverage caveats:** lens-spec not run (SMA-249/229 are prose backlog issues, no AC chain — conformance via the holistic panel + owner live-test); gitleaks not installed (secret scan skipped; Rust src/tests only).

**Decision: ✅ Approve.** Both fixes are correct and verified, and the owner has already live-tested the behavior. **Merge remains owner-gated** — the gate does not merge or tag.

PR #54, opened 2026-06-27
feat: in-file navigation — go to line (`:`) [Phase 1 of 2]

smarzban:
## In-file navigation — Phase 1: **Go to line** (`:`)

First half of the in-file-navigation feature (Linear project *In-file navigation*). **Staged build:** this PR ships **go-to-line only**; in-file search (`/`, `n`/`N`) follows as PR #2.

### What & why
The viewer can move *between* files (tree, and the `f` finder) but had no way to navigate *within* the open file. This adds the universal `:N` primitive: press `:`, type a line number, `Enter` jumps the content pane there.

### Behavior
- `:` opens a one-line **bottom prompt**; digits build the number, non-digits are ignored, `Backspace` edits.
- `Enter` jumps so the target source line is visible (landed near the top); **out-of-range clamps** to the last line; **empty `Enter`** closes with no jump; `Esc` cancels (scroll unchanged).
- **View-gated:** available only in the **source-mapped** (syntax/content) view, where a source line maps 1:1 to a display row. In rendered-markdown / diff / full-diff (or with nothing selected) it shows a one-line *"unavailable"* notice and opens nothing — honest about the absent line→row map (see ADR-0006).
- Fully **read-only** — pure in-memory navigation; no file/git mutation, no persistent state, no event hook (opens only on the explicit `:` key).

### Acceptance criteria covered
- **AC-1** open prompt (source-mapped) · **AC-2** digit entry · **AC-3** jump visible · **AC-4** 1-based, clamp out-of-range · **AC-5** empty-Enter no-op · **AC-6** Esc no-op · **AC-7** unavailable+notice in transformed views (incl. `FullDiff`)
- **AC-21** (go-to-line portion) prompt keys don't fire viewer intents · **AC-N6** no event/auto trigger

### Tasks → commits
| Task | Commit | Advances |
|---|---|---|
| T-1 content `scroll_to_line` | `f3e43dc` | AC-3, AC-4 |
| T-2 `OpenGoToLine` intent + `:` key + view-gated open | `f1ffce3` | AC-1, AC-7, AC-N6 |
| T-3 keystroke + confirm + cancel | `1a3aa6f` | AC-2, AC-3, AC-4, AC-5, AC-6 |
| T-4 event-loop routing + Presenter bottom prompt | `0e0a6e2` | AC-1, AC-7 |
| T-5 pty e2e | `f9f36d1` | AC-21 |
| docs (README/CHANGELOG/ARCHITECTURE) | `1a229a1` | — |

New modules: `src/infile.rs` (prompt-modal state). Reuses go-to-file's `src/prompt.rs` `PromptInput` and the `app.rs` modal raw-key routing.

### Tests
Unit (`intent`, `input`), integration (`controller` — open/gate/clamp/empty/Esc/directory-edge), Presenter (bottom-prompt render + AC-7 notice, **zero snapshot churn**), and a pty **e2e** (`:`+`j`-swallowed+`40`→jump; `:` on markdown→unavailable). **469 tests pass**; clippy `-D` / fmt / release all clean.

### Notes for the reviewer
- **No deferred shortcuts / `SHORTCUT(T-N)` ceilings** recorded by the build.
- This branch's base includes the docs-backfill #53 (go-to-file + tree-title), merged to `main` first so docs stay consistent.
- The full spec chain (Brief / AC / Design / Tech Stack / Plan / gate-report) lives in the Linear project *In-file navigation* and in `specs/in-file-navigation/` (local-only/gitignored); ADR-0006 records the search-over-displayed-content + view-gated-go-to-line decision.
- **Merge is owner-gated** — please don't auto-merge.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -37,6 +37,7 @@ is unit-testable with stubs.
 | `fuzzy` | A pure fuzzy matcher: rank file paths against a typed query (the finder's scoring), no I/O. |
 | `index` | Build the flat, `.gitignore`-aware list of repo file paths the finder searches. |
 | `prompt` | A reusable single-line text-input buffer (push / backspace / clear) backing the finder query — and future keyboard prompts. |
+| `infile` | In-file-navigation modal state: which bottom prompt is open (go-to-line now; in-file search later), its `prompt` input buffer, and the content-scroll snapshot for cancel-restore. |
 | `input` | Map crossterm key events → intents. |
 | `intent` | The closed set of user intents (one exhaustive enum). |
 | `controller` | Orchestrate intents → state changes; hold the ephemeral session state; dispatch renders to the worker; map mouse events (clicks, wheel, divider + scrollbar drags) against the fed-back geometry; on a worktree switch, rebuild the root-bound services through a provider factory and respawn the render worker. |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,11 @@ All notable changes to this project are documented here. The format is based on
 ## [Unreleased]
 
 ### Added
-- **Go to line (`:`).** Press `:` in the syntax/content view to open a prompt and jump the content
-  pane to a source line by number — `Enter` jumps (out-of-range clamps to the last line), `Esc`
-  cancels. In a rendered-markdown or diff view (where a source line has no 1:1 display row) `:` shows
-  a one-line "unavailable" notice instead of opening the prompt. Read-only navigation. (The first
-  half of in-file navigation; in-file search follows.)
+- **Go to line (`:`).** Press `:` to open a prompt and jump the content pane to a source line by
+  number — `Enter` jumps (out-of-range clamps to the last line), `Esc` cancels. Works in every view:
+  in a rendered-markdown or diff view (where a source line has no 1:1 display row) confirming switches
+  the file to the line-numbered content view and jumps there. Read-only navigation. (The first half of
+  in-file navigation; in-file search follows.)
 - **Go to file (`f`).** Open a fuzzy finder over every file in the tree and jump straight to one by
   name — type to filter, `↑` / `↓` to move, `Enter` to open, `Esc` to cancel; `←` / `→` (or the
   horizontal wheel) scroll long result rows, and the result list has a draggable scrollbar.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ All notable changes to this project are documented here. The format is based on
 ## [Unreleased]
 
 ### Added
+- **Go to line (`:`).** Press `:` in the syntax/content view to open a prompt and jump the content
+  pane to a source line by number — `Enter` jumps (out-of-range clamps to the last line), `Esc`
+  cancels. In a rendered-markdown or diff view (where a source line has no 1:1 display row) `:` shows
+  a one-line "unavailable" notice instead of opening the prompt. Read-only navigation. (The first
+  half of in-file navigation; in-file search follows.)
 - **Go to file (`f`).** Open a fuzzy finder over every file in the tree and jump straight to one by
   name — type to filter, `↑` / `↓` to move, `Enter` to open, `Esc` to cancel; `←` / `→` (or the
   horizontal wheel) scroll long result rows, and the result list has a draggable scrollbar.
--- a/README.md
+++ b/README.md
@@ -49,8 +49,9 @@ right into the tree. It opens beside whatever you're doing and never touches you
 - **Jump to any file** — press `f` to open a fuzzy finder over every file in the tree
   (`.gitignore`-aware); type to filter, `↑` / `↓` to move, and `Enter` to open — far faster than
   scrolling the tree in a large repo.
-- **Go to a line** — press `:` and type a line number to jump the content pane straight there
-  (in the syntax/content view); out-of-range clamps to the last line.
+- **Go to a line** — press `:` and type a line number to jump the content pane straight there;
+  in a rendered-markdown or diff view it switches to the line-numbered content view to make the
+  jump. Out-of-range clamps to the last line.
 - **Switch worktree on the fly** — press `W` to re-root the viewer at another git worktree of
   the repo without relaunching; it pre-selects the worktree a herdr agent is working in, so you
   can jump straight to it. Read-only — it changes only *what you're viewing*, never the branch
@@ -116,7 +117,7 @@ you only add the keybinding. The [keys](#keys) are below; deeper detail lives in
 | `v` | Cycle the content view mode |
 | `e` | Open the selected file in `$EDITOR` |
 | `f` | **Go to file** — open a fuzzy finder over every file in the tree; type to filter, `↑` / `↓` move, `Enter` opens the selected file, `Esc` cancels (`←` / `→` scroll long paths) |
-| `:` | **Go to line** — open a prompt and jump the content pane to a source line by number (`Enter` jumps, `Esc` cancels; out-of-range clamps to the last line). Available in the syntax/content view; in a rendered-markdown or diff view it shows a one-line "unavailable" notice |
+| `:` | **Go to line** — open a prompt and jump the content pane to a source line by number (`Enter` jumps, `Esc` cancels; out-of-range clamps to the last line). Works in any view; in a rendered-markdown or diff view, confirming switches to the line-numbered content view and jumps there |
 | `y` | Copy the selected file's **repo-relative** path to the clipboard (e.g. `src/app.rs`) |
 | `Y` | Copy the selected file's **absolute** path to the clipboard |
 | `Tab` | Move focus between the tree and content columns |
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ right into the tree. It opens beside whatever you're doing and never touches you
 - **Jump to any file** — press `f` to open a fuzzy finder over every file in the tree
   (`.gitignore`-aware); type to filter, `↑` / `↓` to move, and `Enter` to open — far faster than
   scrolling the tree in a large repo.
+- **Go to a line** — press `:` and type a line number to jump the content pane straight there
+  (in the syntax/content view); out-of-range clamps to the last line.
 - **Switch worktree on the fly** — press `W` to re-root the viewer at another git worktree of
   the repo without relaunching; it pre-selects the worktree a herdr agent is working in, so you
   can jump straight to it. Read-only — it changes only *what you're viewing*, never the branch
@@ -114,6 +116,7 @@ you only add the keybinding. The [keys](#keys) are below; deeper detail lives in
 | `v` | Cycle the content view mode |
 | `e` | Open the selected file in `$EDITOR` |
 | `f` | **Go to file** — open a fuzzy finder over every file in the tree; type to filter, `↑` / `↓` move, `Enter` opens the selected file, `Esc` cancels (`←` / `→` scroll long paths) |
+| `:` | **Go to line** — open a prompt and jump the content pane to a source line by number (`Enter` jumps, `Esc` cancels; out-of-range clamps to the last line). Available in the syntax/content view; in a rendered-markdown or diff view it shows a one-line "unavailable" notice |
 | `y` | Copy the selected file's **repo-relative** path to the clipboard (e.g. `src/app.rs`) |
 | `Y` | Copy the selected file's **absolute** path to the clipboard |
 | `Tab` | Move focus between the tree and content columns |
@@ -226,7 +229,7 @@ A few things on the way:
 
 - **In-app help overlay** — a `?` key to show every keybinding (and setup tips) without leaving the viewer.
 - **Settings & customization** — a config file for keymaps, the default split, themes, and your own renderer/editor commands.
-- **In-file navigation** — `/` to search within the open file and `:` to jump to a line.
+- **In-file search** — `/` to search within the open file and `n` / `N` to cycle matches (the `:` go-to-line jump has shipped).
 
 **Hit a bug, or want a feature?** Please [open an issue](https://github.com/smarzban/herdr-file-viewer/issues) — bug reports and feature requests are very welcome.
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -157,6 +157,20 @@ fn event_loop(terminal: &mut DefaultTerminal, controller: &mut Controller) -> io
                     }
                     dirty |= fx.redraw;
                 }
+                // While a bottom prompt (go-to-line) is open, route every key press to handle_prompt_key so
+                // digits/printables edit the prompt instead of firing viewer intents (AC-21). Mutually exclusive
+                // with the finder arm above — only one modal is ever open.
+                Event::Key(key) if key.kind == KeyEventKind::Press && controller.prompt_open() => {
+                    let fx = controller.handle_prompt_key(key);
+                    if fx.clear {
+                        let _ = terminal.clear();
+                        dirty = true;
+                    }
+                    if fx.quit {
+                        return Ok(());
+                    }
+                    dirty |= fx.redraw;
+                }
                 Event::Key(key)
                     if key.kind == KeyEventKind::Press
                         && let Some(intent) = input::map_key(key) =>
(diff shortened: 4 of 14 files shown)


smarzban:
## Review Gate — Round 1
🚫 **BLOCK** — 6 blocking finding(s) must be resolved or justified.

Findings: 11 total — 0 critical · 1 high · 6 medium · 4 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-simplify + lens-spec + lens-subtle-correctness + tools · models: codex:gpt-5.5, ollama:glm-5.2:cloud, claude:claude-opus-4-8, ollama:kimi-k2.7-code:cloud, deterministic
⚠️ **Coverage:** 11/14 planned reviewer passes voted — lost: holistic/ollama:kimi-k2.7-code:cloud (non-vote: ollama backend exited 1 (claude.ai connectors disabled — auth-source precedence). Holistic still covered by codex+glm+opus (3 votes).); lens-simplify/ollama:glm-5.2:cloud (non-vote: same ollama auth-source flake. lens-simplify still covered by codex (which surfaced the stale-doc finding).); tools/gitleaks (secret scan skipped — gitleaks not on PATH.)

### Must fix
- 🔴 **[HIGH]** \[correctness\] Mouse events are not gated while the go-to-line prompt is open — `src/controller.rs:942` · 3/4 models _(correctness)_
  handle\_mouse() guards the picker (inert) and finder (routed) but has no \`self.prompt.is\_some()\` guard, while app.rs routes only KEY events to handle\_prompt\_key. So with the go-to-line prompt open, a wheel/click/drag still hits the tree/content beneath: a tree click changes the selection, then Enter's auto-switch reads the NEW \`tree.selected()\` and either jumps in the wrong file or — if the new selection is a directory — inserts a bogus SyntaxContent override on a directory path and sets pending\_goto against a render that is never dispatched (clear\_content sends no job), leaving stale state. This breaks the modal invariant the picker/finder uphold.
  _Fix:_ Add a guard at the top of handle\_mouse mirroring the picker: \`if self.prompt.is\_some() { return Effects::noop(); }\` (the go-to-line prompt is keyboard-only), placed alongside the existing picker/finder gates.
- 🟠 **[MEDIUM]** \[correctness\] scroll\_to\_line targets the wrong row when content wrap is on — `src/controller.rs:1405` · 2/4 models _(correctness)_
  scroll\_to\_line sets content\_scroll = line-1 and clamps against rendered\_line\_count(), assuming a source line maps 1:1 to a display row. That holds only when wrap is off. wrap\_for (line 787) forces wrapping for every file — including SyntaxContent — when the \`w\` wrap\_override is on, so a long source line spans multiple display rows. Under wrap\_override, go-to-line N lands on display row N-1 (mid-wrapped-line), not source line N, and the clamp uses the wrapped row total rather than the source-line count, so out-of-range clamping is also wrong. The bug is silent: the user believes they are on line N.
  _Fix:_ When wrap is in effect, compute the display-row offset of the target source line by summing wrapped\_rows of the preceding lines (rendered\_line\_count\_for already does this per-line) instead of using line-1; or disable go-to-line / document the no-wrap requirement. At minimum correct the doc comment's 1:1 claim.
- 🟠 **[MEDIUM]** \[correctness\] Go-to-line can miss wrapped source lines — `src/controller.rs:1408` · 1/4 models _(correctness)_
  AC-3 requires Enter on line N to make source line N visible, but this maps the source line directly to display row N-1. If wrapping is enabled and earlier source lines occupy multiple display rows, the viewport can land above the requested source line and leave it off-screen.
  _Fix:_ When wrapping is active, compute the target display row by summing the wrapped row counts for the source lines before N using the same wrapping logic as the scroll clamp, then clamp that row offset.
- 🟠 **[MEDIUM]** \[concurrency\] Queue source-mode jumps until the current render is applied — `src/controller.rs:1905` · 1/4 models _(concurrency)_
  The SyntaxContent path calls scroll\_to\_line immediately, but after startup or a selection change the selected file's off-thread render may still be in flight. If the current content is empty or from a shorter previous file, the jump clamps against stale content and the later render is applied without redoing the jump.
  _Fix:_ Track the applied render seq/path and, when the latest render for the selected file has not landed, queue pending\_goto against latest\_seq just like the transformed-view path; add a test that confirms :N before the render result arrives.
- 🟠 **[MEDIUM]** \[concurrency\] Go-to-line can clamp against stale content — `src/controller.rs:1907` · 1/4 models _(concurrency)_
  Interleaving: navigating to a source file dispatches render seq N but leaves the previous file in self.content; if the user immediately opens ':' and presses Enter before seq N is polled, scroll\_to\_line computes rendered\_line\_count from the old content. When seq N later lands, poll replaces the content without reapplying the requested line, so the jump silently lands at the wrong offset, often the top.
  _Fix:_ Track the seq/path of the content currently applied, and when latest\_seq is not yet applied, queue pending\_goto against latest\_seq even for SyntaxContent; then apply the jump in poll when that render result lands.
- 🟠 **[MEDIUM]** \[concurrency\] In-flight auto-switch jump clobbers a newer synchronous go-to-line confirm — `src/controller.rs:1907` · 1/4 models _(concurrency)_
  After a first \`:\`-confirm in a markdown/diff view sets \`overrides\[A\]=SyntaxContent\`, dispatches render seq N, and queues \`pending\_goto=(N, line1)\`, \`selected\_view\_mode()\` immediately reports SyntaxContent (it reads the override) even though the source render hasn't landed. If the user reopens \`:\` and confirms a second line before render N arrives, the Enter handler takes the synchronous branch (line 1905-1907) and calls \`scroll\_to\_line(line2)\` — but it never clears \`pending\_goto\`. When render N finally lands, \`poll()\` matches \`pending\_goto.seq == latest\_seq\` and overwrites the scroll with \`scroll\_to\_line(line1)\`. The user typed line2 but lands on line1. The synchronous path assumes no auto-switch jump is in flight, an ordering that isn't guaranteed; the window widens on large files / slow renderers.
  _Fix:_ Clear \`self.pending\_goto = None;\` at the top of the non-empty Enter branch (or in the synchronous \`scroll\_to\_line\` arm) so a newer confirm always supersedes a still-in-flight auto-switch jump — mirroring how \`dispatch\_render()\` clears it for navigation.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[maintainability\] Prompt is not torn down on re-root, despite the field doc claiming it is — `src/controller.rs:245` · 1/4 models _(maintainability)_
  The prompt field doc says it is "Closed by confirm/cancel (T-3) and by a re-root / new render," but re\_root closes picker and finder (lines 480, 486) yet never sets self.prompt = None, and dispatch\_render clears only pending\_goto, not self.prompt. Today this is unreachable because re\_root/dispatch\_render are modal-gated while the prompt is open, but the finder was given the same symmetric teardown explicitly for forward-looking safety (line 481-485), so the prompt should match — and the doc is currently false.
  _Fix:_ Add \`self.prompt = None;\` in re\_root alongside the picker/finder clears, and fix the field doc to describe only the real close paths (confirm/cancel).
- ⚪ **[LOW]** \[maintainability\] Intent::OpenGoToLine doc contradicts the implemented AC-7 behavior — `src/intent.rs:71` · 3/4 models _(maintainability)_
  The doc says go-to-line is "Available only in a source-mapped (syntax/content) view; in a transformed view it shows an unavailable notice and opens nothing (AC-7)." But open\_go\_to\_line opens the prompt for any selected file regardless of view mode (the revised AC-7), and the only unavailable notice fires for a directory. The comment describes the superseded design and will mislead future readers.
  _Fix:_ Rewrite the doc to match the revised AC-7: \`:\` opens the prompt for any file; confirming in a transformed view auto-switches to SyntaxContent and queues the jump; the unavailable notice fires only when no file is selected.
- ⚪ **[LOW]** \[test-coverage\] No test exercises go-to-line with content wrap enabled (AC-3) — `tests/controller.rs:661` · 1/4 models _(test-coverage)_
  Every go-to-line jump test runs with wrap off, so the exact case where the line-to-row mapping is incorrect (wrap on, long lines above the target) is untested; no test would fail if AC-3 were violated under wrap.
  _Fix:_ Add a test that toggles wrap on, renders content with lines that wrap, jumps to a line below a wrapped line, and asserts the target source line is within \[content\_scroll, content\_scroll+height).
- ⚪ **[LOW]** \[test-coverage\] Presenter test and e2e comments still describe the removed 'unavailable in markdown/diff' behavior — `tests/presenter.rs:2394` · 1/4 models _(test-coverage)_
  unavailable\_notice\_renders\_for\_ac7 feeds a hand-crafted notice string and asserts it renders, but after the AC-7 revision the controller never emits an 'unavailable in this view' notice for a transformed view (the only real notice is the directory/no-file case). The test name and the e2e function name (\`...\_is\_unavailable\_in\_markdown\`) describe pre-revision behavior, so they no longer guard the path they claim to and could mask a regression.
  _Fix:_ Rename/retarget the test to assert the real AC-7 notice path (directory selected → notice, prompt stays closed) and the auto-switch path, and refresh the now-stale 'unavailable' wording in the e2e/test comments.

- **[MEDIUM]** \[concurrency\] Preserve pending go-to-line across status refresh redispatch — `src/controller.rs:2216`
  _Dismissed:_ Verified in poll() (src/controller.rs \~2200-2225): the re-root status branch calls dispatch\_render() unconditionally, which clears pending\_goto. Real but advisory, not a merge-blocker: (1) the window is a re-root's async status fetch STILL in flight AND the user confirming ':N' in a transformed view inside that sub-second — re-root status completes in ms; (2) the outcome is a fail-safe DROPPED jump (read-only, no crash, user re-presses ':'), never a wrong-file/wrong-line jump; (3) a naive carry-forward is unsafe — the status redispatch re-renders with the freshly-landed changed-set, which can flip the selected file content-\>Diff, so re-applying the old source-line jump in a now-Diff view would be wrong; the current drop is the safer behavior. Logged as a follow-up (Linear) rather than fixed under the anti-over-engineering directive.

smarzban:

I ran the full discovery panel on this PR: **holistic ×4** (codex/glm/opus + a kimi non-vote) plus four lenses fired by trigger — **lens-spec** (revised AC-1..7/21/N1-3/N6 passed in-prompt, since the spec is gitignored), **lens-security** (digit-buffer input parsing), **lens-simplify** (dead-code / over-engineering — requested given the major auto-switch change), and **lens-subtle-correctness** (the off-thread `pending_goto` seq-keyed jump). The deterministic scan ran clean. Coverage: 11/14 planned passes voted — two ollama runs lost to the known auth-precedence non-vote (holistic still had 3 votes; lens-simplify still had codex), and gitleaks isn't installed (secret scan skipped). All recorded in the Coverage line above.

I read the code behind every gating cluster before deciding. What I'm **fixing this round**:

- 🔴 **Mouse not gated while the prompt is open (HIGH, 5 models).** Confirmed: `handle_mouse` guards picker + finder + Shift but has **no `prompt` guard**, while `app.rs` routes only *key* events to the prompt. A mid-prompt click/wheel changes the tree selection, so Enter's auto-switch then acts on the **wrong file** — or, on a directory, writes a bogus `SyntaxContent` override + an orphaned `pending_goto` against a render that never dispatches. Fix: make the mouse inert while the prompt is open, mirroring the picker.
- 🟠 **`scroll_to_line` wrong under wrap (MEDIUM, 4 models).** Confirmed: `wrap_for` forces wrapping for **every** file (incl. SyntaxContent) when the `w` override is on, so `content_scroll = line-1` lands on display row N-1, not source line N — AC-3 silently fails under wrap. Fix: map the source line to its cumulative wrapped-row offset when `effective_wrap()`, clamp the source line to the source-line count. + a wrap test.
- 🟠 **Concurrency: stale-content & double-confirm races (MEDIUM).** Confirmed by tracing: `selected_view_mode()` reads the *override*, so it reports SyntaxContent before the switch render lands — the synchronous jump then runs against stale content, and a second confirm gets clobbered by the first's queued `pending_goto`. Fix: track an `applied_seq` and only take the synchronous path when the latest render has actually been applied; otherwise queue the jump (and clear any superseded pending jump on a new confirm).
- ⚪ **Cheap correctness/hygiene I'm folding in** (advisory but real): the stale `intent.rs` AC-7 doc (3 models), `re_root` not clearing `self.prompt` (its field doc claims it does), and the stale "unavailable-in-markdown" test/e2e names.

What I **dismissed** (code-checked): the *status-refresh redispatch drops `pending_goto`* medium — real but a narrow window (a re-root status fetch in flight ∩ a transformed-view confirm), fail-safe (a dropped jump, never a wrong-file jump), and a naive carry-forward would be *unsafe* because the redispatch can flip the view to Diff. I'll log it as a follow-up rather than fix it under the over-engineering directive.

**Decision: 🔴 Request changes.** Implementing the fixes above now; Round 2 will verify them and check for regressions.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ holistic + tools · models: codex:gpt-5.5, deterministic
⚠️ **Coverage:** 2/4 planned reviewer passes voted — lost: holistic-verify/claude:claude-opus-4-8 (non-vote: opus returned prose with no parseable JSON array on the verification prompt (both fg and bg). Verification covered by codex gpt-5.5 (reliable parser, 0 findings).); tools/gitleaks (secret scan skipped — gitleaks not on PATH.)

### Progress since Round 1
✅ Resolved (6): \[correctness\] Mouse events are not gated while the go-to-line prompt is open; \[correctness\] scroll\_to\_line targets the wrong row when content wrap is on; \[correctness\] Go-to-line can miss wrapped source lines; \[concurrency\] Queue source-mode jumps until the current render is applied; \[concurrency\] Go-to-line can clamp against stale content; \[concurrency\] In-flight auto-switch jump clobbers a newer synchronous go-to-line confirm — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

Round 2 was single-model **verification** (not a fresh discovery sweep): for each Round 1 finding, is it resolved at HEAD (`c654cd1`), and did any fix introduce a direct regression? Verified by **codex gpt-5.5** (reliable parser) → **0 findings**, plus the deterministic scan (clean) and my own code re-read. Opus non-voted on the verification prompt (prose, no JSON array) — lost coverage, surfaced in the Coverage line, never treated as a pass.

### What this PR implements
Phase 1 of in-file navigation: **go-to-line** (`:`). `:` opens a bottom prompt for any selected file; digits build a 1-based line number; Enter jumps the content pane to that source line (clamped); Esc/empty-Enter close without moving. In a transformed view (markdown/diff) confirming **auto-switches** the file to the source-mapped content view and jumps once it re-renders (revised AC-7). The prompt is modal — keyboard-only, no viewer intents fire while it's open.

### Fixed (cumulative across both rounds)
- 🔴 **HIGH** — Mouse is now inert while the prompt is open (`handle_mouse` prompt guard); a mid-prompt click/wheel can no longer change the selection and send a confirm to the wrong file or strand a directory override.
- 🟠 **MEDIUM** — `scroll_to_line` is wrap-aware: under the `w` override it maps the source line to its cumulative wrapped-row offset (shared `wrapped_rows_before` helper) instead of `line-1`, so `:N` reaches source line N under wrap (AC-3).
- 🟠 **MEDIUM** — Async-jump hardening: a new `applied_seq` guard takes the synchronous jump only when the latest render has landed (otherwise queues against the in-flight render), and each confirm clears `pending_goto` so a newer confirm supersedes an in-flight auto-switch jump. No more stale-content clamp or clobbered confirm.
- ⚪ **LOW** — Refreshed the stale `Intent::OpenGoToLine` doc to the revised AC-7; `re_root` now clears `self.prompt` (field doc accurate); retargeted the stale "unavailable" presenter test + e2e name to the real behaviour.
- **Coverage:** +4 regression tests (mouse-modal, wrap-jump, stale-content race, double-confirm race). Suite: **474 green**, clippy `-D` / fmt / release clean.

### Deferred / not covered
- **Status-refresh redispatch dropping a queued jump** (Round 1 medium, dismissed): a narrow, fail-safe window (re-root status fetch in flight ∩ a transformed-view confirm) where a naive carry-forward would itself be unsafe (the redispatch can flip the view to Diff). Logged as a follow-up, not a merge-blocker.
- **gitleaks** not installed → the deterministic secret scan was skipped both rounds (no secrets expected in this change).
- Opus verification non-vote (Round 2) — verification carried by codex.

**Decision: ✅ Approve.** The gate verdict is `pass` with no real blocker deferred. Per this repo's owner-gated merge policy, I'm **not** merging — the green, reviewed PR is yours to merge when you're ready.

PR #55, opened 2026-06-27
feat: in-file search (`/`, `n`/`N`) — Phase 2 of in-file navigation

smarzban:
## In-file search (`/`, `n`/`N`) — Phase 2 of in-file navigation

The second (and final) half of the in-file-navigation feature. Phase 1 (go-to-line, `:`) shipped in
**#54**; this PR adds **search within the open file**. Read-only navigation throughout — it only
highlights and scrolls; no file or git mutation, no persistent state, zero new Cargo deps.

### What it does

- **`/`** opens a bottom search prompt (`Search: <query>`) — for the selected file, in **every** view
  mode (rendered-markdown, diff, syntax/content). It searches the content **as displayed**, so a
  string visible in the pane is found regardless of how it's rendered.
- Every match **highlights as you type**; the content scrolls to the first match. Matching is
  **smartcase** (a lowercase query is case-insensitive; any uppercase letter makes it case-sensitive)
  and **literal** (regex metacharacters match literally, never interpreted).
- **`Enter`** commits the search — the prompt closes, highlights persist, and a status bar shows the
  match count + hints: `Search: foo (3/12) · n next · N prev · Esc clear`.
- **`n` / `N`** cycle to the next / previous match in document order, wrapping at the ends with a
  notice. The **current** match is visually distinct (yellow) from the others (cyan).
- **`Esc` / `q`** exits the search (clears the highlight), layered before unzoom/close.
- The search clears automatically when the displayed content changes (a different file, view mode, or
  diff baseline).
- `/` requires a file selected (a notice otherwise, mirroring `:`), and when only the tree is showing
  it zooms the selected file so the matches are visible.

### Acceptance criteria

Search portion of the in-file-navigation contract (the spec chain is gitignored / local-only per the
repo convention — criteria reproduced here for review):

- **AC-8** `/` opens a one-line search prompt at the bottom, in any view mode.
- **AC-9** every occurrence of the query in the displayed content is highlighted.
- **AC-10** when ≥1 match exists, the content scrolls so a match is visible.
- **AC-11** the current match is visually distinguished from the other matches.
- **AC-12** smartcase matching.
- **AC-13** finds text as displayed in every view mode.
- **AC-14** `Enter` commits — prompt closes, query/matches/highlight persist.
- **AC-15** `n` next / `N` previous, scrolling the new current match into view.
- **AC-16** match navigation wraps at the ends, with a notice.
- **AC-17** `Esc` while the prompt is open cancels — clears highlight, restores the pre-open scroll.
- **AC-18** a no-match query highlights nothing and doesn't move the scroll.
- **AC-19** `n`/`N` are inert without a committed search with ≥1 match.
- **AC-20** persisted highlighting clears on a new search or when the displayed content changes.
- **AC-21** while a prompt is open, printable keys edit it and don't trigger viewer actions.
- **AC-22** incremental re-search **and re-highlight** stay within the 300 ms budget at the size cap.
- **AC-23** search covers only the content shown; a size-cap-truncated file's beyond-boundary text isn't matched.
- **AC-N1/N2** no filesystem or git mutation. **AC-N3** no persistent state. **AC-N4** matches only the
  open file's displayed content. **AC-N5** literal substring, never a regex. **AC-N6** acts only via the
  explicit `/` `n` `N` keys — no event hook (the manifest declares no `[[events]]`).

> **AC-8 refinement** (mirrors Phase 1's AC-7): `/` opens for a **selected file** in any view; on a
> directory / nothing it shows a notice. (Owner UX, folded in after a live test — see below.)

### Tasks (one atomic, reviewed, green commit each)

| Task | Commit | Advances |
|---|---|---|
| T-6 Search Matcher (`src/search.rs`) | `34f120e` | AC-9, 12, 18, N4, N5 |
| T-7 Match Highlighter (`src/highlight.rs`) | `31d5703` | AC-9, 11 |
| T-8 Intents + keys (`/` `n` `N`) | `0799c71` | AC-8, 19, N6 |
| T-9 Incremental matching | `bac5de2` | AC-9, 10, 18 |
| T-10 Commit + `n`/`N` + wrap | `eb72a68` | AC-14, 15, 16, 19 |
| T-11 Cancel + clear lifecycle | `f472ef7` | AC-17, 20 |
| T-12 Presenter highlight + ViewState | `bbaf191` | AC-8, 9, 11 |
| T-13 All-modes / smartcase / truncation + e2e | `96482d8` | AC-12, 13, 21, 23 |
| T-14 Performance | `1958e1b` | AC-22 |
| T-15 Negative criteria & conformance | `1c0dac0` | AC-N1, N2, N3, N4, N6 |
| docs | `62a53ca` | — |
| Search UX revisions (owner live-test feedback) | `8cb2408` | AC-8 refinement + UX |

### Notes for the reviewer

- **T-14 caught a real AC-22 violation:** `highlight::apply` was O(lines × matches) — 2,845 ms on a
  5,000-line file with ~25k matches. Fixed to ~33 ms (group matches by line once + reconstruct each
  line's plain text once per line), behavior-preserving (all T-7 tests stay green). Perf guard added.
- **Search reads the *displayed* content** (`content.lines` plain text), not the source file — that's
  why it's view-agnostic and why it's the second half of in-file navigation (go-to-line is
  source-mapped-view only).
- **Highlighting composes over the delegated render** (ADR-0006) — it re-segments the existing ANSI
  spans, never re-tokenizes.
- **No `SHORTCUT` ceilings.** 557 tests green; clippy `-D warnings` / fmt / release all clean.

Staged as PR #2 of the in-file-navigation feature (Linear project *In-file navigation*, SMA-239…248).


smarzban:
## Review Gate — Round 1

Findings: 10 total — 0 critical · 0 high · 1 medium · 8 low · 1 info.
_Reviewed by:_ holistic-codex + holistic-glm + holistic-kimi + holistic-opus + lens-security + lens-simplify + lens-spec + lens-subtle-correctness + tools · models: codex:gpt-5.5, ollama:kimi-k2.7-code:cloud, ollama:glm-5.2:cloud, claude:claude-opus-4-8, deterministic
⚠️ **Coverage:** 9/10 planned reviewer passes voted — lost: tools/gitleaks (gitleaks not on PATH — committed-secret scan skipped)

### Must fix
- 🟠 **[MEDIUM]** \[concurrency\] In-flight renders can leave search state bound to stale content — `src/controller.rs:2519` · 4/4 models _(concurrency)_
  A render dispatch clears the old search immediately, but if the user opens/types/commits a search before that render result is applied, \`poll\` replaces \`self.content\` without clearing or recomputing the new search state. The retained match offsets and counts then refer to the previous displayed content and can highlight/navigate the wrong lines in the newly rendered file.
  _Fix:_ When applying a render result in \`poll\`, clear committed search state, and if a Search prompt is currently open recompute \`refresh\_search()\` against the newly applied content before redrawing.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[correctness\] open\_search zooms based on content\_width == 0, which is also true before the first draw — `src/controller.rs:1992` · 1/4 models _(correctness)_
  \`content\_width\` is initialized to 0 and only set after the first frame is drawn. The \`open\_search\` zoom check treats \`content\_width == 0\` as 'content pane not visible (narrow layout)', but it also fires before any draw—e.g. in direct controller tests, or if the startup order ever changes—forcing zoom on a wide terminal.
  _Fix:_ Distinguish 'narrow layout' from 'not yet drawn' by tracking whether the content pane has ever been laid out, or by checking the actual last-known frame width against the narrow threshold instead of relying solely on \`content\_width == 0\`.
- ⚪ **[LOW]** \[correctness\] Empty searches can be committed as active no-match searches — `src/controller.rs:2149` · 2/4 models _(correctness)_
  If the user types then backspaces to an empty query and presses Enter, \`self.search\` remains \`Some\` with an empty query, leaving a persistent \`Search: (no matches)\` status and causing the next Esc/q to clear that empty search instead of performing the normal close action.
  _Fix:_ On Enter, clear \`self.search\` when the prompt query is empty, or avoid creating \`SearchState\` for an empty query in \`refresh\_search\`.
- ⚪ **[LOW]** \[maintainability\] Dead \`line\` binding in \`next\_match\` requires \`let \_ = line\` suppression — `src/controller.rs:2183` · 1/4 models _(maintainability)_
  \`next\_match\` destructures the current match's \`line\` into the tuple \`(len, current, line)\`, never uses it (it re-fetches \`next\_line\` after the match), and then has to write \`let \_ = line; // suppress unused warning\`. The sibling \`prev\_match\` does the identical job without extracting that field. The extra binding plus suppression comment is noise a reader must untangle to confirm nothing depends on it.
  _Fix:_ Drop \`line\` from the match arm tuple — \`(s.matches.len(), s.current)\` — and delete the \`let \_ = line;\` line, mirroring \`prev\_match\`.
- ⚪ **[LOW]** \[performance\] highlight::apply re-segments the entire content (not just the visible window) on every redraw while a search is active — `src/presenter.rs:544` · 1/4 models _(performance)_
  Whenever state.search is Some, draw\_content runs highlight::apply over ALL of state.content.lines on every redraw — including scroll/zoom/resize keystrokes, not just the search keystrokes. AC-22's guard caps testing at 5k lines/300ms; beyond that cap (no file-line limit is enforced yet — SMA-252) a committed search makes ordinary scrolling re-segment the whole document each keypress, scaling linearly with file size.
  _Fix:_ Advisory only — within the team's stated 300ms budget at the 5k cap. If/when SMA-252's line limit lands, or if large-file lag is observed, restrict highlight::apply to the visible row window (content\_scroll .. content\_scroll+content\_height) rather than the full line vector.
- ⚪ **[LOW]** \[correctness\] smartcase rule is documented as 'any uppercase letter' but only checks ASCII uppercase — `src/search.rs:39` · 1/4 models _(correctness)_
  The module comment says a query with 'any uppercase letter' is case-sensitive, yet \`find\_matches\` uses \`c.is\_ascii\_uppercase()\` to decide. A non-ASCII uppercase query like 'É' therefore takes the case-insensitive path while still not matching 'é' (because the fold is ASCII-only), so the documented rule and implementation disagree.
  _Fix:_ Either update the module-level doc/test assertions to say 'any uppercase ASCII letter', or switch the check to \`c.is\_uppercase()\` and explicitly document that folding is ASCII-only so non-ASCII uppercase queries behave predictably.
- ⚪ **[LOW]** \[maintainability\] Manual byte-walk loop reinvents \`str::match\_indices\` — `src/search.rs:53` · 1/4 models _(maintainability)_
  Both \`find\_matches\_case\_sensitive\` and \`find\_matches\_case\_insensitive\` hand-roll a \`while search\_start \<= line.len().saturating\_sub(query.len()) { line\[start..\].find(query) ... }\` loop to collect non-overlapping byte offsets — exactly what \`str::match\_indices(query)\` returns (disjoint matches, byte start offsets). The code comment even states "\`str::match\_indices\` yields byte offsets directly" yet doesn't use it. The hand-rolled saturating-sub/offset arithmetic is more to verify for off-by-one safety than the std iterator, and it duplicates the same loop across two near-identical functions.
  _Fix:_ Replace the loop with \`for (start, m) in haystack.match\_indices(needle) { matches.push(Match { line: line\_idx, start, end: start + m.len() }); }\`. The case-sensitive/insensitive variants then collapse to one function that picks \`needle\`/\`haystack\` (folded or not) up front, removing the duplicated body.
- ⚪ **[LOW]** \[test-coverage\] AC-20 baseline-toggle clears committed search is untested — `tests/controller.rs:6037` · 1/4 models _(test-coverage)_
  AC-20 explicitly lists '(file/view/baseline)' as displayed-content changes that must clear a committed search. Dedicated tests exist for the new-search, file-select, and cycle-view triggers, but none pairs Intent::ToggleBaseline with a committed search. A regression dropping the search-clear from the baseline path (toggle\_baseline -\> dispatch\_render at src/controller.rs:1701) would leave the other two tests green and ship stale highlights against the re-rendered diff.
  _Fix:_ Add a sibling to content\_change\_clears\_committed\_search\_cycle\_view that commits a search in a git-backed controller, handles Intent::ToggleBaseline, and asserts ctrl.search() is None.
- ⚪ **[LOW]** \[test-coverage\] AC-22 perf guard exercises the 5000-line cap but not the 1 MB byte cap or the combined per-keystroke budget — `tests/search\_perf.rs:29` · 1/4 models _(test-coverage)_
  AC-22 bounds incremental re-search AND re-highlight at the size cap '\<=5000 lines / 1 MB'. build\_lines() produces 5000 x \~72 bytes (\~360 KB), so the byte dimension (\~1 MB, \~2.7x the bytes the per-line to\_ascii\_lowercase allocates each keystroke) is not exercised, and find\_matches + highlight::apply are only timed separately, never as the combined per-keystroke path the criterion describes.
  _Fix:_ Add a case at \~1 MB (e.g. \~5000 lines of \~200 chars) and time find\_matches followed by highlight::apply together against the 300 ms budget to cover the combined incremental path.
- ⚪ **[INFO]** \[maintainability\] \`validate\_match\_plain\` takes both \`plain\` and \`line\_text\_len\` when one derives the other — `src/highlight.rs:104` · 1/4 models _(maintainability)_
  \`line\_text\_len\` is computed in \`apply\` as the sum of span content lengths and \`plain\` is the concatenation of those same span contents, so \`plain.len() == line\_text\_len\` always. Passing both as separate parameters invites a reader to wonder whether they can ever diverge (they can't), and is a redundant argument.
  _Fix:_ Drop the \`line\_text\_len\` parameter and use \`plain.len()\` inside \`validate\_match\_plain\` (or drop \`plain\` and pass only the length where the boundary check needs it). Either removes the dual-source-of-truth.

smarzban:
## Orchestrator review — Round 1 · 🔴 Request changes

I ran the gate with holistic on all four lineages (codex `gpt-5.5`, kimi, glm, opus) and the lenses
on the non-gpt models (lens-spec → opus, lens-security → glm, lens-subtle-correctness → kimi,
lens-simplify → opus), plus the deterministic scan. **9/10 passes voted** (gitleaks absent → the
committed-secret scan was skipped; the rest of the deterministic scan ran clean — no conflict markers,
focused tests, or committed artifacts).

**Agreeing with the BLOCK.** One real finding, and I confirmed it in the code:

- 🟠 **MEDIUM — in-flight render leaves search state bound to stale content** (`controller.rs:2519`,
  flagged by all 4 holistic models). T-11 clears `self.search` at the *top of `dispatch_render`*
  (dispatch time), but the actual content swap happens later in `poll()` when the off-thread render
  result lands — and `poll()` sets `self.content = result.content` **without** clearing `self.search`.
  So a search opened/committed in the window between dispatch and apply (e.g. `/`-searching a file
  right after selecting it, before its render lands) keeps matches computed against the *old* content
  and overlays them on the *new* content. The stated AC-20 invariant ("any displayed-content change
  wipes a committed search") is false for the poll path. **Fixing:** clear `self.search` when a render
  is applied in `poll()`, and recompute `refresh_search()` if a Search prompt is open.

**Also folding in** (advisory, but cheap and worth it — several from the simplify lens):
the dead `let _ = line` in `next_match`; an empty-query Enter that commits a phantom `(no matches)`
search; the hand-rolled byte loop in `search.rs` (→ `str::match_indices`, collapsing the two
near-identical fns); the redundant `line_text_len` param on `validate_match_plain`; the smartcase doc
vs `is_ascii_uppercase`; and a missing AC-20 baseline-toggle clear test.

**Dismissed (verified not-real / out of budget):**
- *zoom-before-first-draw* — the event loop draws before it reads input, so `content_width` is set
  before the first `/`; the `content_width == 0` check is the same pattern the go-to-file finder uses
  for the tree-only layout.
- *highlight re-segments whole content* — measured at ~33 ms over the 5k-line cap (T-14), inside the
  300 ms budget; a visible-window optimization is a follow-up (relates to SMA-252), not a blocker.
- *perf guard 1 MB byte-cap* — the line cap is the dominant cost dimension and is exercised with a
  many-match worst case; a byte-cap case is a nice-to-have follow-up.

**Decision: 🔴 Request changes.** Fixing the MEDIUM + the folded advisories, then a verification round.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ tools + verify-r2 · models: ollama:kimi-k2.7-code:cloud, deterministic
⚠️ **Coverage:** 2/4 planned reviewer passes voted — lost: verify-r2/claude:claude-opus-4-8 (unparseable output (prose non-vote) — re-verified by kimi); tools/gitleaks (gitleaks not on PATH — committed-secret scan skipped)

### Progress since Round 1
✅ Resolved (1): \[concurrency\] In-flight renders can leave search state bound to stale content — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:
## Orchestrator review — Round 2 · ✅ Approve

Round-2 **verification** (single non-gpt model, per the gate's converge-don't-rediscover design):
kimi `kimi-k2.7-code:cloud` checked each Round-1 finding against the fix diff + the deterministic
scan re-ran. **0 still-open, 0 regressed.** (The opus verifier pure-prosed a non-vote — surfaced in
the Coverage line, NOT treated as a pass; re-verified by kimi, which is why it's covered.)

### What this PR implements
In-file search for the open file: `/` opens a labelled prompt (`Search: <query> (i/n)`) for a selected
file in **every** view mode; matches highlight as you type (current match distinct), `Enter` commits
with an `n next · N prev · Esc clear` status bar, `n`/`N` cycle with wrap, `Esc`/`q` exit. Smartcase +
literal (never regex), over the content **as displayed**, read-only, zero new deps. Phase 2 (final
half) of in-file navigation; Phase 1 (go-to-line) shipped in #54.

### Fixed (cumulative)
- **MEDIUM (R1, all 4 holistic models)** — `poll()` left a committed/in-flight search bound to stale
  content after an in-flight render landed; now reconciled at the content swap (recompute if a Search
  prompt is open, else clear), with a red-then-green regression test. This is the AC-20 invariant
  holding for the async poll path, not just the dispatch path.
- Folded advisories: phantom empty-query commit; `search.rs` → `str::match_indices` (two case fns
  collapsed to one); redundant `validate_match_plain` param; dead `next_match` binding; smartcase doc;
  + an AC-20 baseline-toggle regression test.

### Deferred / not covered
- *Visible-window-only highlighting* — current full-content re-segmentation is ~33 ms at the 5k-line
  cap (within the 300 ms budget); a windowing optimization is a follow-up, tied to the file-line-limit
  work (SMA-252).
- A 1 MB byte-cap perf case (line cap is the dominant dimension and is exercised).
- Committed-secret scan skipped (gitleaks not installed) — the rest of the deterministic scan ran clean.

### Verdict
560 tests green; clippy `-D warnings` / fmt / release clean. **Decision: ✅ Approve.** Merge is
owner-gated — handing the reviewed, passing PR to the owner.

PR #59, opened 2026-06-30
Optimisations — M1–M5 (audit-driven tech-debt)

smarzban:
## Optimisations — M1–M5 (audit-driven tech-debt)

Clears the M1–M5 milestones of the **Optimisations** backlog (the latest whole-repo audit at
`53874c4`): quick wins, test-timing robustness, test-coverage depth, UX papercuts, and
accessibility. Twenty issues, one milestone per commit, plus a final cleanup commit.

**Scope note:** M6 (the `controller.rs` god-module decomposition) is intentionally **not** in this
PR — it's a dedicated structural refactor for a later PR. This PR is read-only-preserving,
additive, and behavior-compatible.

### What's in it (by milestone)

**M1 · Quick wins**
- Extract the duplicated `wait_bounded` subprocess reaper (`render.rs` + `update/mod.rs`) into one
  shared `src/proc.rs` — pure dedup, total wall-clock bound unchanged.
- Doc corrections: `release.yml` prebuilt-gate comment (now version-match, not commit-exact),
  `ARCHITECTURE.md` (the `update-check.json` on-disk exception + the new `proc` module),
  `docs/install.md` (git as a runtime requirement; `NO_UPDATE_CHECK` any-value), `docs/renderers.md`
  (bundled palette; glow-is-Go cargo-fallback correction), `herdr-plugin.toml` pane comment.

**M2 · Test timing**
- New `perf` cargo feature gates the absolute-stopwatch budgets (`render_perf`, `tree_perf`, the
  `reroot` AC-17 budget) off the default PR lane (`cargo test --features perf` runs them).
- `search_perf` + `index_perf` rewritten as **relative-scaling** asserts (`time(2N) < ~4×time(N)`,
  with a minimum-base floor) so they catch an O(n²) regression without flaking on a slow/loaded
  runner — these stay on the default lane.
- pty e2e tests now `expectrl` wait-for-content instead of fixed `thread::sleep`-then-read
  (eliminates the torn-read flake class).

**M3 · Test coverage**
- `no_handled_intent_mutates_the_filesystem` strengthened to route **every** `Intent::ALL` variant
  through the real handler (closing any modal it opens first) against a content-aware FS/git
  snapshot — the read-only invariant (AC-N1/N2) is now genuinely exercised.
- A modal × intent cross-product guard matrix (5 modal states × every `Intent::ALL` variant) —
  modal isolation (AC-5/AC-6), no second modal, tree/FS unchanged; auto-covers new intents.
- Table-driven git porcelain/diff parser tests (malformed/truncated/invalid-UTF8/rename-copy/
  unknown codes + direct `classify` assertions).
- OSC-52 clipboard-exfiltration ingestion test (AC-27) on the renderer path; CLI smoke gated off
  the update network.

**M4 · UX papercuts**
- Content pane no longer shows a stale file under a new title while an off-thread render is in
  flight — `Rendering…` placeholder + title-switches-with-body + supersession.
- `RendererError` enum → clean fallback notices (no raw OS errno).
- `EditorOutcome`/`SpawnError` enums → distinguish a launch failure from a non-zero editor exit.
- A `? help` hint on the content border + empty-state guidance for blank/zero-match panes.

**M5 · Accessibility**
- Non-color cues alongside color-only signalling: tree status glyphs (`M`/`A`/`D`/`?`/`●`), active
  help-tab `▶ ` marker, `(current)` worktree label, current-match `REVERSED+BOLD`, and
  theme-relative (`REVERSED`) status/prompt bars.
- Tree horizontal scroll is now keyboard-reachable: `H`/`L` (Shift+`h`/`l`), focus-gated, clamped
  to the measured max — closes the AC-18 keyboard-first gap.

**Final commit · build-process-ref sweep**
- Removed every internal tracker reference (issue IDs, plan task IDs, review notes, audit tags)
  from `src/` + `tests/` comments and the public CHANGELOG; kept `AC-N`/`ADR-NNNN`. Comment/string
  only — zero code change.

### Decisions to flag (per the issue ACs)

- **SMA-337 (macOS e2e):** the two macOS `#[ignore]` e2e tests (`e2e_help`, `e2e_editor`) are
  **retained** with their rationale. The expectrl change removed the assertion-site sleeps, but the
  unavoidable no-content-anchor sleeps (terminal-resume settle, Esc-close with no screen overwrite)
  remain — so the macOS flake risk they document remains. Re-enabling can only be proven on the
  macOS CI matrix; kept ignored until then.
- **SMA-334 (perf CI job):** deliberately **no** CI perf job added. The relative-scaling tests
  already guard O(n²) on every PR; running machine-dependent absolute budgets on a shared runner —
  even advisory — is noise that matches this repo's own "don't hard-assert wall-clock on shared CI"
  lesson. The perf budgets run on demand via `cargo test --features perf`.

### Verification

Green bar on Linux: `cargo build --release` · `cargo test` · `cargo clippy --all-targets -- -D
warnings` · `cargo fmt --check` — all clean. Perf-gated lane (`cargo test --features perf`) green.
The full CI matrix (incl. macOS) runs on this PR.


smarzban:
## Review Gate — Round 1

Findings: 4 total — 0 critical · 0 high · 2 medium · 2 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-subtle-correctness · models: codex:gpt-5.5, ollama:glm-5.2:cloud
⚠️ **Coverage:** 6/8 planned reviewer passes voted — lost: holistic/ollama:kimi-k2.7-code:cloud (holistic/ollama:kimi-k2.7-code:cloud: timed out after 600000ms); holistic/claude:claude-opus-4-8 (holistic/claude:claude-opus-4-8: unparseable output)
🔍 **Scan tier degraded:** secrets: gitleaks not on PATH — skipped (install gitleaks to enable secret scanning)

### Must fix
- 🟠 **[MEDIUM]** \[concurrency\] content pane strands on "Rendering…" if the render worker dies — `src/controller.rs:2921` · 1/2 models · _by:_ holistic, lens-subtle-correctness _(concurrency)_
  dispatch\_render now wipes content to "Rendering…" and sets content\_rendering=true BEFORE the job send. The send's failure is swallowed (\`let \_ = self.job\_tx.send(...)\`), and content\_rendering is only ever cleared by poll applying a matching result or by clear\_content on a directory/empty selection. If the worker thread is gone — e.g. git.diff (line 503) panics inside the worker, which is NOT wrapped in catch\_unwind unlike content.render — no result ever arrives, the channel is Disconnected, and every subsequent file dispatch re-sets the placeholder; the pane is stuck on "Rendering…" with content\_rendering=true for the rest of the session. The adjacent comment ('the pane keeps its last content rather than panicking') is now stale — this PR's wipe contradicts it. Pre-PR the pane genuinely kept its last content on send failure.
  _Fix:_ On send failure, revert the optimistic state: if \`self.job\_tx.send(...).is\_err()\`, clear content\_rendering and either restore the prior content or fall through to clear\_content so the pane degrades gracefully instead of freezing on the placeholder. (Equivalently: only commit the placeholder/flag after a successful send, or wrap git.diff in the existing catch\_unwind.)

### Advisory (non-blocking)
- ⚪ **[LOW]** \[error-handling\] Resume failure hides editor changes — `src/app.rs:337` · 2/2 models · _by:_ holistic _(error-handling)_
  If the editor launches and exits successfully but \`resume\_tui()\` fails, \`LiveEditor\` returns \`NotLaunched\`; the controller's \`NotLaunched\` branch does not refresh git state or re-render content, so edits made by the editor can remain stale.
  _Fix:_ Represent this as a terminal-restore failure after takeover, or otherwise route it through a controller path that refreshes git state/content while still showing the restore error notice.
- ⚪ **[LOW]** \[time\] Update probe can spend two timeout budgets — `src/update/mod.rs:155` · 2/2 models · _by:_ lens-subtle-correctness _(time)_
  \`recv\_timeout(PROBE\_TIMEOUT)\` can consume almost the full 10s waiting for stdout, then the success branch gives \`wait\_bounded\` another full \`PROBE\_TIMEOUT\`. If \`git ls-remote\` closes stdout near the deadline and then hangs before exit, the documented hard wall-clock bound becomes roughly 20s.
  _Fix:_ Create a single \`Instant\` deadline before \`recv\_timeout\` and pass \`deadline.saturating\_duration\_since(Instant::now())\` to \`wait\_bounded\`, matching the renderer call-site.

- **[MEDIUM]** \[correctness\] Git support now silently requires newer Git — `src/git.rs:255`
  _Dismissed:_ Pre-existing on main, NOT a regression of this PR: git.rs:234 on main carries the identical \`--attr-source={EMPTY\_TREE}\` arg and \`git diff main..HEAD\` shows zero changes to it. It is a deliberate read-only security hardening (read attributes from the empty tree so a repo-planted filter=/diff= driver cannot execute during a read-only query). Out of scope for an M1-M5 verdict; valid as a separate maintainer follow-up (probe/cache --attr-source support, or document a min Git version).

smarzban:

Panel: **glm + codex** on holistic, plus **lens-security** and **lens-subtle-correctness** on both (subprocess reaping, untrusted git-output parsing, and the async-render concurrency are the sensitive surfaces here). `kimi` timed out and `opus` returned prose on the holistic pass twice each — both are surfaced in the Coverage line above, so this is a 2-lineage panel (bolstered by 4 lens shots); I'm treating the verdict as moderate-confidence accordingly. **lens-security came back clean from both models.**

**Agree — fixing:**
- 🟠 **controller stuck-placeholder** (escalated to MEDIUM by the subtle-correctness lens): a real worker-death strand introduced by the M4 loading state.
- 🔵 **editor-resume → `NotLaunched`** (LOW, 2 models): a real edge case from M4/SMA-344 — a resume failure after a successful edit skipped the git refresh and showed a false "Could not open editor".
- 🔵 **update double-timeout** (LOW, 2 models): real; the probe could spend ~2× its bound.

**Dismissed (code-checked):** the git `--attr-source` MEDIUM is **pre-existing on `main`** (git.rs:234, identical, unchanged by this PR's diff) and is a deliberate read-only security hardening — not an M1–M5 regression. Logged for a separate maintainer follow-up (probe/cache `--attr-source`, or document a min Git version).

**Decision: 🔴 Request changes** — fixing the three real findings, then verifying.

smarzban:
## Review Gate — Round 2

_Reviewed by:_ verify · models: codex:gpt-5.5
🔍 **Scan tier degraded:** secrets: gitleaks not on PATH — skipped (install gitleaks to enable secret scanning)

### Progress since Round 1
✅ Resolved (1): \[concurrency\] content pane strands on "Rendering…" if the render worker dies — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

All three real Round-1 findings are fixed and verified resolved at HEAD (`487e0c8`) with no regression; the deterministic scan is clean and **the full CI matrix (both ubuntu + both macOS, clippy, audit) is green**.

**Fixed (cumulative):**
- **controller stuck-placeholder** — the `Rendering…` placeholder is now committed only when the worker dispatch succeeds, **and** the off-thread `git.diff` read was moved inside the worker's `catch_unwind`. (Round-1's verify initially passed on the partial fix; on re-reading the finding I caught that `git.diff` — the trigger the reviewer actually named — still sat outside the panic guard, so I completed the fix and re-verified rather than sign off on a partial resolution.)
- **editor-resume** — a terminal-restore failure after a successful editor run now returns `TookOver` (git refresh + re-render + full repaint), not `NotLaunched`.
- **update double-timeout** — a single combined deadline now bounds the probe, matching the renderer call-site and honoring `proc::wait_bounded`'s contract (completing SMA-328's reaper unification).

**Also in this PR (beyond the panel):** a CI-only pty-e2e flake (an `expect` content-anchor that raced the no-renderer fallback re-render under parallel CI) was root-caused and de-flaked; CI went green.

**Deferred / not covered:** git `--attr-source` min-version (pre-existing, dismissed above); `kimi`/`opus` holistic lineages (lost to timeout/non-vote — see the Coverage line). M6 (the `controller.rs` decomposition) is intentionally out of scope for this PR.

**Decision: ✅ Approve** — agrees with the `pass` verdict. Holding the merge for the repo owner's sign-off (owner-gated).

PR #78, opened 2026-07-08
Mouse text selection: content copy in line-select mode, plus ambient drag-select

### LINKED ISSUE #76: cant select file contents in ghostty
While using herdr and ghostty I can't select any file contents, or even go into selection mode.

Image

### LINKED ISSUE #80: Shift + Mouse going across panes
Having this issue in Mac: 

Image

When I am trying to use the text selector, it start copying across the panes within Herdr

j-pollack:
## Summary

Two related improvements to mouse text selection in the content pane.

### 1. Line-select mode copies content (and supports mouse selection)
- Adds click-drag mouse text selection in line-select mode: press to start, drag to extend (auto-scrolling past an edge), release to finish, with the selection highlighted live as you drag.
- `Enter`/`y`/`Y` now copies the actual selected content instead of a `path:line` reference — the joined line text for a keyboard selection, or the exact dragged character span for a mouse selection. The `bat --style=numbers` line-number gutter is stripped, indentation is preserved, and stray control bytes are removed before hitting the clipboard (tabs kept).
- Confirmation notice names what was copied (e.g. "Copied lines 42-58" / "Copied selection").
- A mouse press now places the selection caret; a plain click no longer drops a line marker, and double-click no longer copies.

### 2. Ambient drag-select (no mode required)
- Click-drag over the content pane during **normal navigation** — no mode to enter — selects text character-by-character; releasing the drag **auto-copies** it to the clipboard ("Copied selection"), leaving the highlight standing for feedback.
- Implemented via a `content_selection` field held **outside** the `Modal` enum, so `Modal::None` stays in force and every keyboard binding keeps its normal meaning (that's what makes it ambient, not a mode). It reuses the line-select char primitives verbatim; **L line-select mode is left untouched** and keyboard-driven, and the two are mutually exclusive by construction.
- Drawn with a gutter-less overlay (no ▶/│ glyph, no content shift) so it reads as a plain text selection.
- Cleared on any content change, `Esc`, a click-away, a plain click, or entering L mode; it survives content scrolling. Guards: `Shift`+drag stays reserved for the terminal's native selection; copying is blocked over a non-file selection and while a render is in flight.

### Both
- `Shift`+mouse is left alone so the terminal's own native selection still works.
- No changes to `intent.rs`, `input.rs`, or `app.rs`.

Closes #76

## Test plan
- [x] `cargo test` — unit + integration coverage in `tests/lineselect.rs` (line-select + ambient: auto-copy on release, plain-click focuses, click-away/Esc/content-change clears, Shift inert, directory/render-placeholder guards, no leak into L mode) and `tests/presenter.rs` (gutter-less char highlight, no content shift). One pre-existing e2e pty test (`search_routes_keys_…`) is flaky on macOS and fails identically on the base branch.
- [x] `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all --check` clean under the CI-pinned toolchain (1.96.0).
- [x] Manually verified the line-select mouse selection/copy in a real terminal (ghostty, per #76).
- [x] Manual spin of the **ambient** drag-select in a real terminal still pending — the ambient path is exercised by the automated mouse-handler tests above, but has not yet been driven live.

### No mode
image

### Line selection mode

image

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -43,7 +43,7 @@ is unit-testable with stubs.
 | `text_layout` | A pure text-wrapping helper: how many display rows a line occupies at a given width — shared by the content pane, the finder, and the help overlay. No I/O. |
 | `prompt` | A reusable single-line text-input buffer (push / backspace / clear) backing the finder query — and future keyboard prompts. |
 | `infile` | In-file-navigation modal state: which bottom prompt is open (go-to-line or in-file search), its `prompt` input buffer, the live `SearchState` (query, matches, current match), and the content-scroll snapshot for cancel-restore. |
-| `lineselect` | Line-select modal state (copy-line-reference): anchor + marker source-line indices, the focus-gated `L` entry that auto-switches to the source view, key/mouse handling, and formatting the `path:line` / `path:start-end` reference for the clipboard. Read-only. |
+| `lineselect` | Line-select modal state (copy-line-content): anchor + marker source-line indices, the focus-gated `L` entry that auto-switches to the source view, key/mouse handling, and extracting the selected lines' text (joined by newlines, residual control bytes stripped) for the clipboard. Read-only. |
 | `help` | Help overlay state: the embedded changelog source and About text, plus the section and vertical scroll position for the `?` overlay. Pure; no I/O — the changelog is compiled in at build time. |
 | `input` | Map crossterm key events → intents. |
 | `intent` | The closed set of user intents (one exhaustive enum). |
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -43,7 +43,7 @@ is unit-testable with stubs.
 | `text_layout` | A pure text-wrapping helper: how many display rows a line occupies at a given width — shared by the content pane, the finder, and the help overlay. No I/O. |
 | `prompt` | A reusable single-line text-input buffer (push / backspace / clear) backing the finder query — and future keyboard prompts. |
 | `infile` | In-file-navigation modal state: which bottom prompt is open (go-to-line or in-file search), its `prompt` input buffer, the live `SearchState` (query, matches, current match), and the content-scroll snapshot for cancel-restore. |
-| `lineselect` | Line-select modal state (copy-line-content): anchor + marker source-line indices, the focus-gated `L` entry that auto-switches to the source view, key/mouse handling, and extracting the selected lines' text (joined by newlines, residual control bytes stripped) for the clipboard. Read-only. |
+| `lineselect` | Line-select modal state: anchor + marker source-line indices (plus mouse char carets), the focus-gated `L` entry that auto-switches to the source view, key/mouse handling, and the two confirms — formatting the `path:line` / `path:start-end` reference (`Enter`) and extracting the selected text (`y`/`Y`, joined by newlines, gutter stripped, residual control bytes removed) for the clipboard. Read-only. |
 | `help` | Help overlay state: the embedded changelog source and About text, plus the section and vertical scroll position for the `?` overlay. Pure; no I/O — the changelog is compiled in at build time. |
 | `input` | Map crossterm key events → intents. |
 | `intent` | The closed set of user intents (one exhaustive enum). |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,10 @@ All notable changes to this project are documented here. The format is based on
 - **Line-select mode copies content with `y`/`Y`.** One selection, two products: `Enter` keeps
   copying the `path:line` / `path:start-end` **reference** (unchanged since 1.9.0), while the new
   `y`/`Y` copy the selected **content** — for a keyboard selection, the line(s) joined by
-  newlines; for a mouse selection, the exact character span. The syntax view's line-number gutter
-  (`bat --style=numbers`) is stripped from the content copy, indentation is preserved, and
-  residual control bytes are removed before it reaches the clipboard (tabs kept). The confirmation
+  newlines; for a mouse selection, the exact character span. Whole-line copies come from the
+  file's own retained text, byte-faithful (real tabs, no renderer decoration); the syntax view's
+  line-number gutter (`bat --style=numbers`) never reaches the copy, indentation is preserved,
+  and residual control bytes are removed before it reaches the clipboard (tabs kept). The confirmation
   notice names what was copied (e.g. "Copied lines 42-58" / "Copied selection").
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,21 @@ All notable changes to this project are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+- **Mouse text selection in line-select mode.** Click-drag in the content pane to select text
+  character-by-character (press at the start, drag to the end — the pane auto-scrolls past an
+  edge — release), with the selected characters highlighted as you drag; `Enter`/`y`/`Y` copies
+  exactly the dragged characters. `Shift`+mouse is left untouched so the terminal's own native
+  selection still works.
+
 ### Changed
-- **Line-select mode now copies the selected lines' content, not a `path:line` reference.** `Enter`,
-  `y`/`Y`, or a double-click copies the actual text of the selected line (or, for a range, every
-  line joined by newlines, indentation preserved) — so it pastes as real code rather than a
-  location string. The syntax view's line-number gutter (`bat --style=numbers`) is stripped from
-  the copy, and residual control bytes are removed before it reaches the clipboard (tabs kept).
-  The confirmation notice names the copied line range (e.g. "Copied lines 42-58"). Marker
-  movement, extend, mouse handling, and the source-view auto-switch are unchanged.
+- **Line-select mode now copies the selected content, not a `path:line` reference.** `Enter`,
+  `y`/`Y` copies the actual text: for a keyboard selection, the line(s) joined by newlines; for a
+  mouse selection, the exact character span. The syntax view's line-number gutter
+  (`bat --style=numbers`) is stripped from the copy, indentation is preserved, and residual control
+  bytes are removed before it reaches the clipboard (tabs kept). The confirmation notice names what
+  was copied (e.g. "Copied lines 42-58" / "Copied selection"). A mouse press now places the
+  selection caret (previously a click placed the line marker; a double-click no longer copies).
 
 ## [1.10.0] - 2026-07-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,18 +16,19 @@ All notable changes to this project are documented here. The format is based on
   selection still works, and dragging over placeholder text (a directory / empty pane) copies nothing.
 - **Mouse text selection in line-select mode.** Click-drag in the content pane to select text
   character-by-character (press at the start, drag to the end — the pane auto-scrolls past an
-  edge — release), with the selected characters highlighted as you drag; `Enter`/`y`/`Y` copies
-  exactly the dragged characters. `Shift`+mouse is left untouched so the terminal's own native
-  selection still works.
+  edge — release), with the selected characters highlighted as you drag. `Shift`+mouse is left
+  untouched so the terminal's own native selection still works.
+- **Line-select mode copies content with `y`/`Y`.** One selection, two products: `Enter` keeps
+  copying the `path:line` / `path:start-end` **reference** (unchanged since 1.9.0), while the new
+  `y`/`Y` copy the selected **content** — for a keyboard selection, the line(s) joined by
+  newlines; for a mouse selection, the exact character span. The syntax view's line-number gutter
+  (`bat --style=numbers`) is stripped from the content copy, indentation is preserved, and
+  residual control bytes are removed before it reaches the clipboard (tabs kept). The confirmation
+  notice names what was copied (e.g. "Copied lines 42-58" / "Copied selection").
 
 ### Changed
-- **Line-select mode now copies the selected content, not a `path:line` reference.** `Enter`,
-  `y`/`Y` copies the actual text: for a keyboard selection, the line(s) joined by newlines; for a
-  mouse selection, the exact character span. The syntax view's line-number gutter
-  (`bat --style=numbers`) is stripped from the copy, indentation is preserved, and residual control
-  bytes are removed before it reaches the clipboard (tabs kept). The confirmation notice names what
-  was copied (e.g. "Copied lines 42-58" / "Copied selection"). A mouse press now places the
-  selection caret (previously a click placed the line marker; a double-click no longer copies).
+- **In line-select mode, a mouse press now places the selection caret** (previously a click placed
+  the line marker, and a double-click copied — copying is now always an explicit `Enter`/`y`).
 
 ## [1.10.0] - 2026-07-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ All notable changes to this project are documented here. The format is based on
 ## [Unreleased]
 
 ### Added
+- **Ambient mouse text selection in the content pane.** Click-drag over the content pane during
+  normal navigation — no mode to enter — to select text character-by-character (press at the start,
+  drag to the end while the pane auto-scrolls past an edge, release). Releasing the drag copies the
+  selected text to the clipboard automatically ("Copied selection"), and the highlight stays for
+  feedback; `Esc`, a click elsewhere, or switching files clears it. Line-select mode (`L`) remains a
+  separate keyboard-driven mode. `Shift`+drag is left untouched so the terminal's own native
+  selection still works, and dragging over placeholder text (a directory / empty pane) copies nothing.
 - **Mouse text selection in line-select mode.** Click-drag in the content pane to select text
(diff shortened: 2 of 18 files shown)


bnivanov:
Yes please, I was struggling with this today!

smarzban:
Thanks @j-pollack, well-built PR. I've pushed three small changes:

- **013804b**: brought back Enter = copy the `path:line` reference (that's the shipped v1.9 behavior people rely on for pasting into agent chats), and moved your content copy to `y`/`Y`. One selection, two products: drag with the mouse, then `y` for the text or Enter for the reference.
- **80f4dea**: copies now come from the retained raw source lines instead of parsing bat's output. Real tabs survive, no gutter stripping needed, and the line-number heuristic is just a fallback (it could eat a line that starts with its own number).
- **2728603**: made selection work under wrap (markdown wraps by default, so this mattered a lot in practice). The wrap row math was a rough simulation of ratatui's wrapper and drifted on real code lines, so clicks landed a line off. It's now a faithful port of WordWrapper, with a test that renders through a real Paragraph and asserts the breaks match, so it can't silently drift again.

The core design here is all yours and it held up great. Thanks again.

My review gate runs next, then merge.

smarzban:
## Empanel — Round 1

Findings: 12 total — 0 critical · 0 high · 3 medium · 9 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-subtle-correctness + lens-tests · models: opencode-go:deepseek-v4-pro, ollama:glm-5.2, openai-codex:gpt-5.5, anthropic:claude-opus-4-8
⚠️ **Coverage:** 9/10 planned reviewer passes voted — lost: lens-spec (no in-repo spec/AC exists for this external PR)
⚠️ **Reviewer warnings:** holistic/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns)); lens-subtle-correctness/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns)); lens-tests/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns))

### Must fix
- 🟠 **[MEDIUM]** \[correctness\] Gutter heuristic strips real text in source-less views — `src/controller/lineselect.rs:288` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(correctness)_
  gutter\_len\_of falls back to strip\_line\_gutter for every view without retained source, including rendered markdown/plain-text ambient selections that have no line-number gutter. A real line such as line 1 text \`1 Introduction\` (or line 2 text \`2 Things\`) is treated as a guttered line and copied/highlighted without its leading number, violating the advertised exact text selection for wrapped/prose views.
  _Fix:_ Only apply the gutter-stripping heuristic for views/render results known to have a renderer gutter (for example SyntaxContent fallback, and any explicit full-diff gutter case), or carry explicit gutter/source metadata in RenderResult; return gutter 0 for rendered markdown/plain text without such metadata.
  _Possible sibling (co-located — verify same defect):_ \[correctness\] Character selection is misaligned in line-select mode with wrap on (the overlay glyph shifts the renderer's wrapping but not the caret math) — `src/controller/lineselect.rs:300`
- 🟠 **[MEDIUM]** \[correctness\] Wrapped L-mode mouse mapping ignores the added marker glyph — `src/controller/lineselect.rs:346` · 2/4 models · _by:_ holistic _(correctness)_
  In line-select mode the presenter prepends a one-column marker/bar glyph to every content line before ratatui wraps it, but char\_at\_content\_col computes wrap row starts using the original line width/content. For wrapped selections near a wrap boundary, the screen row/column can therefore map to the wrong source line or character, causing the newly added mouse selection to highlight/copy text different from what the user dragged over.
  _Fix:_ Make the controller's wrapped row/column mapping use the same text/width the presenter wraps in L mode (e.g. include the overlay glyph in wrap calculations or reserve a separate non-wrapping gutter and reduce the text width consistently), and add a regression test for a line whose wrap changes after the glyph is added.
- 🟠 **[MEDIUM]** \[test-coverage\] autoscroll\_selection (drag-past-edge auto-scroll) is exercised by no test — `src/controller/mouse.rs:97` · 1/4 models · possible same-defect siblings — combined 3/4 · _by:_ lens-tests _(test-coverage)_
  \`autoscroll\_selection\` nudges \`content\_scroll\` when a selection drag runs above the top (\`row \< c.y\`) or below the bottom (\`row \>= c.y + c.height\`) of the content viewport, for BOTH the L-mode drag (line 78) and the ambient drag (line 188). This is a documented, user-facing behavior (CHANGELOG/README: "the pane auto-scrolls past an edge"). No test in tests/lineselect.rs drives a Drag event past the viewport edge and asserts \`content\_scroll\` changed, so a refactor that breaks the boundary checks or the \`scroll\_content(±1)\` calls would silently regress with no test failure. Every existing drag test keeps the cursor inside the viewport (content height 20, drag within rows 1-4).
  _Fix:_ Add a test that sets a short content viewport (e.g. height 4) over a 20-line body, begins an ambient (and an L-mode) drag, then sends a Drag event with \`row\` past the bottom edge, and asserts \`content\_scroll\` increments by 1; symmetrically for a drag above \`content\_inner.y\`.
  _Possible sibling (shared topic):_ \[correctness\] Character selection is misaligned in line-select mode with wrap on (the overlay glyph shifts the renderer's wrapping but not the caret math) — `src/controller/lineselect.rs:300`
  _Possible sibling (shared topic):_ \[correctness\] Gutter-strip heuristic can silently drop a leading "N " from a copied selection in transformed/non-source views — `src/controller/lineselect.rs:36`
  _Possible sibling (shared topic):_ \[concurrency/state\] drag field not cleared on re-root — `src/controller/mod.rs:1680`

### Advisory (non-blocking)
- ⚪ **[LOW]** \[correctness\] Character selection is misaligned in line-select mode with wrap on (the overlay glyph shifts the renderer's wrapping but not the caret math) — `src/controller/lineselect.rs:300` · 1/4 models · possible same-defect siblings — combined 3/4 · _by:_ holistic _(correctness)_
  In L mode the presenter prepends a 1-column ▶/│ glyph to each source line \*before\* ratatui wraps it, so the real wrap boundaries are those of "glyph+text". But \`char\_at\_content\_col\` computes \`wrap\_row\_starts\` on the glyph-less \`text\`, and the vertical row accounting (\`rendered\_line\_count\_for\`/\`content\_row\_of\_line\`) also omits the glyph. On continuation rows of a wrapped line the mapped caret is off by the glyph width, so the highlighted/copied character span is misaligned. Only affects L mode + \`w\` wrap toggle (a supported combo per the README); the ambient path (no glyph) is correct.
  _Fix:_ Account for the active overlay glyph column when simulating wrap in L mode — run the wrap simulation against the same glyph-prefixed text the presenter draws (or offset the row-start math by the glyph width on the first wrapped row), so the caret matches the rendered wrapping.
  _Possible sibling (co-located — verify same defect):_ \[correctness\] Gutter heuristic strips real text in source-less views — `src/controller/lineselect.rs:288`
  _Possible sibling (shared topic):_ \[test-coverage\] autoscroll\_selection (drag-past-edge auto-scroll) is exercised by no test — `src/controller/mouse.rs:97`
- ⚪ **[LOW]** \[correctness\] Gutter-strip heuristic can silently drop a leading "N " from a copied selection in transformed/non-source views — `src/controller/lineselect.rs:36` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(correctness)_
  For ambient (and line-select fallback) copies over views where no raw source is retained (rendered markdown / diffs, or providers without \`source\`), \`gutter\_len\_of\` falls back to \`strip\_line\_gutter\`, which strips a leading run of digits equal to the displayed line's 1-based index when followed by a space/bar. A displayed line that legitimately begins with that same number + separator (e.g. display line 5 rendering the text "5 apples") gets its "5 " removed from the clipboard content — silent corruption of copied data. Source-mapped SyntaxContent is safe (exact anchoring), so this only bites transformed views, but the copy is unattended (auto-copies to the OS clipboard on drag release).
  _Fix:_ For the ambient/char-selection copy path, only strip a gutter when the content is actually source-mapped (i.e. \`content\_source.is\_some()\` / exact-anchored), and skip the digit heuristic entirely for transformed views where no gutter exists; or gate the heuristic on the syntax view mode.
  _Possible sibling (shared topic):_ \[test-coverage\] autoscroll\_selection (drag-past-edge auto-scroll) is exercised by no test — `src/controller/mouse.rs:97`
- ⚪ **[LOW]** \[test-coverage\] copy\_content\_selection (ambient auto-copy) clipboard-failure path untested — `src/controller/lineselect.rs:385` · 1/4 models · _by:_ lens-tests _(test-coverage)_
  The ambient drag release calls \`copy\_content\_selection\`, whose error arm sets \`format!("Could not copy selection: {e}")\`. No test exercises an ambient drag-release against a failing clipboard — all ambient drag tests use the recording clipboard. The failure-notice wording for the ambient path (distinct from the reference path) and the "leave the highlight standing" behavior on error are not asserted.
  _Fix:_ Add a test with \`FailingClipboard\` that performs an ambient press→drag→release of a non-collapsed span and asserts the notice contains "Could not copy selection" and that the selection highlight is unaffected.
- ⚪ **[LOW]** \[test-coverage\] copy\_line\_content (y) clipboard-failure notice and mode-close-on-error untested — `src/controller/lineselect.rs:578` · 1/4 models · _by:_ lens-tests _(test-coverage)_
  \`clipboard\_error\_shows\_failure\_notice\_no\_panic\` only drives \`Enter\` (the \`copy\_line\_reference\` path, notice "Could not copy line reference: {e}"). The new \`y\`/\`Y\` content-copy path \`copy\_line\_content\` builds a different notice \`format!("Could not copy {label}: {e}")\` and unconditionally calls \`exit\_line\_select()\` after the clipboard match. Neither the failure notice wording nor the mode-close-on-error is exercised for the \`y\` path; a regression that drops \`exit\_line\_select()\` from the error arm (or mislabels the notice) would not be caught.
  _Fix:_ Add a test using \`FailingClipboard\` that enters line-select, presses \`y\`, and asserts the notice contains "Could not copy" AND \`!ctrl.line\_select\_active()\` (the mode closes on error).
- ⚪ **[LOW]** \[test-coverage\] Shift+Y (uppercase 'Y') content-copy spelling is not tested — `src/controller/lineselect.rs:638` · 1/4 models · _by:_ lens-tests _(test-coverage)_
  The handler accepts \`KeyCode::Char('y') \| KeyCode::Char('Y')\`, and the doc explicitly notes \`Y\` arrives as the uppercase char + SHIFT and must be accepted "so holding Shift never makes copy silently fail." Every \`y\`-copy test uses \`key(KeyCode::Char('y'))\` (lowercase, no modifier); none sends \`Char('Y')\`. The \`shift\_j\_extends\_selection\` test covers the \`J\` uppercase spelling for move/extend, but the \`Y\` copy spelling — including that the modifier guard permits it — has no direct guard.
  _Fix:_ Add an assertion that \`handle\_line\_select\_key(key(KeyCode::Char('Y')))\` copies the content and closes the mode, mirroring the lowercase \`y\_copies\_single\_line\_content\` test.
- ⚪ **[LOW]** \[correctness\] wide-glyph column mapping is not implemented — `src/controller/lineselect.rs:80` · 1/4 models · _by:_ holistic _(correctness)_
  char\_index\_at\_col uses col.min(s.chars().count()), assuming one display column per char. For CJK characters and emoji (which occupy 2 terminal columns), this maps a screen column to the wrong character index — e.g. a column of 5 could land on char index 4 instead of 3 if an earlier wide glyph consumed 2 columns. This means mouse-drag selections over wide-glyph text will select the wrong characters.
  _Fix:_ Replace the body of char\_index\_at\_col with a width-summing walk (e.g. using unicode-width) that accounts for each char's display width, as noted in the doc comment. The signature does not need to change.
- ⚪ **[LOW]** \[security\] Bidi override characters not filtered when copying file content to clipboard — `src/controller/lineselect.rs:82` · 1/4 models · _by:_ lens-security _(security)_
  The new \`filter\_control\_keep\_tabs\` function (and the pre-existing \`sanitize\_control\`) use \`char::is\_control()\` which covers C0/C1 control characters (U+0000–U+001F, U+007F, U+0080–U+009F) but does NOT catch Unicode bidi override / format characters (general category Cf, e.g. U+202A–U+202E, U+2066–U+2069). A malicious source file containing these invisible characters would have them copied intact to the clipboard, enabling a 'Trojan Source' attack: when the user pastes the copied code into an editor, the bidi characters can make the code's logical order differ from its displayed order, hiding malicious logic. The PR extends the clipboard-attack surface from paths (where bidi spoofing is less impactful) to actual file content (where it matters more). The fix is to also strip Unicode format characters (general category Cf), or at minimum the bidi override subset.
  _Fix:_ Extend \`filter\_control\_keep\_tabs\` (and \`sanitize\_control\` in text\_layout.rs) to also strip Unicode bidi override/format characters — e.g. filter out any \`char\` whose \`general\_category()\` is \`Cf\` (Format), or explicitly strip the known bidi control set: U+202A–U+202E, U+2066–U+2069, U+200E–U+200F, U+061C.
- ⚪ **[LOW]** \[concurrency/state\] drag field not cleared on re-root — `src/controller/mod.rs:1680` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(correctness)_
  re\_root() explicitly clears modal, last\_click, content\_path, action\_notice, etc., but does not clear self.drag. A stale Drag::ContentSelect from an ambient selection can persist across a re-root. While the next mouse event handles it harmlessly (content\_selection is cleared by dispatch\_render, making the drag handler inert), the state is inconsistent with the other explicit clears in re\_root.
  _Fix:_ Add self.drag = None; in re\_root() alongside the existing self.last\_click = None; and self.modal = Modal::None; clears, so the drag state is reset consistently with other navigation state.
  _Possible sibling (shared topic):_ \[test-coverage\] autoscroll\_selection (drag-past-edge auto-scroll) is exercised by no test — `src/controller/mouse.rs:97`
- ⚪ **[LOW]** \[correctness\] Ambient drag over non-file content leaves an uncopyable highlight — `src/controller/mouse.rs:142` · 1/4 models · _by:_ holistic _(correctness)_
  The ambient press handler seeds \`content\_selection\` for any content hit where \`!content\_rendering\` — including a directory/empty-tree guidance body (a non-file node). On release, \`copy\_content\_selection\` gates on \`is\_file\` and returns early WITHOUT clearing \`content\_selection\`, so the dragged highlight persists over guidance text that can never be copied, with no 'Copied selection' notice. It only clears on Esc / click-away / file-switch. The comment acknowledges the is-file guard doesn't stop the drag, but the resulting lingering highlight is a confusing UX gap.
  _Fix:_ Either skip seeding \`content\_selection\` when the selected node isn't a file (mirror the \`content\_rendering\` early-return with an \`is\_file\` check), or clear \`content\_selection\` in \`copy\_content\_selection\`'s non-file/empty branch so no uncopyable highlight is left standing.

smarzban:
## Empanel — Round 2

Findings: 1 total — 0 critical · 0 high · 1 medium · 0 low · 0 info.
_Reviewed by:_ verify · models: ollama:glm-5.2
⚠️ **Reviewer warnings:** verify/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns))
✅ **Dismissal audit:** all 1 dismissal justification(s) held against the code.

### Progress since Round 1
✅ Resolved (2): \[correctness\] Gutter heuristic strips real text in source-less views; \[correctness\] Wrapped L-mode mouse mapping ignores the added marker glyph — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

- **[MEDIUM]** \[test-coverage\] autoscroll\_selection (drag-past-edge auto-scroll) is exercised by no test — `src/controller/mouse.rs:97`
  _Dismissed:_ Verifier false-negative (absence claim). Spot-checked at HEAD 27d9eda: the test drag\_past\_the\_viewport\_edge\_autoscrolls\_the\_content exists in tests/lineselect.rs:1876; it seeds an ambient drag then drags to a row below the content\_inner bottom edge and above the top edge, asserting content\_scroll increments, then decrements, and stays put on an in-viewport drag — i.e. it drives a drag past both viewport edges. \`cargo test --test lineselect drag\_past\_the\_viewport\_edge\` passes green. The verifier failed to find the test; the finding is resolved.

smarzban:
Ran the multi-model review gate over the whole PR (the ambient feature plus my three maintainer commits). Round 1 flagged three real issues; all are fixed in one commit, and round 2 verified them resolved with no regression.

**Fixed this round:**
- L-mode caret drift under wrap: the ▶/│ overlay glyph takes a column on the first display row, so text wraps one column narrower in L mode, but the caret math simulated full width. Now the wrap simulation accounts for the glyph (a no-op outside L mode, so ambient drag and the scroll math are byte-identical). Continuation-row clicks in L mode + wrap now land on the right character.
- Gutter over-strip in source-less views: a line-number gutter only exists behind the syntax view, which always retains source. `gutter_len_of` now returns 0 when there is no source, so a markdown/diff/plain line that happens to start with its own number keeps it on copy.
- Added tests: source-less ambient copy keeps a leading number, and drag-past-edge autoscroll (which had no coverage).

**Not changed (advisory, tracked as follow-ups):** wide-glyph (CJK/emoji) column mapping stays a documented v1 limitation; bidi/format-character scrubbing on copy; a couple of clipboard-failure test-coverage gaps; an inert stale-drag-on-reroot and an uncopyable highlight over non-file content. None block.

813 tests green, fmt and clippy clean.

Decision: ✅ Approve.

PR #82, opened 2026-07-08
fix: left gap between border and text in markdown/diff views

smarzban:
## What

The rendered-markdown (glow) and diff (delta) views drew their delegate output starting at column 0, so the text hugged the content pane's left border. Syntax-highlighted files already get a small gap from `bat`'s line-number gutter; markdown and diff had none, so they looked inconsistent.

This insets those two transformed views one column from the left border, matching the gap the other files already have.

**Before / after** (zoomed content pane):

```
before (flush)          after (padded)
┌main.rs──────┐         ┌main.rs──────┐
│fn main() {  │         │ fn main() { │
│}            │         │ }           │
└─────? help──┘         └─────? help──┘
```

## How

- New `content_pad_left: bool` on `ViewState`, set by the controller from the **displayed** content's effective mode (`RenderedMarkdown` / `Diff` / `FullDiff`). Syntax/plain files stay flush so they don't get a double gap.
- The pad is applied through a shared `content_block` helper used by **both** `draw_content` and `geometry`, so the drawn text rect and the mouse hit-test geometry inset the border in lockstep (a mismatch would map a content-pane click one column off).

Purely visual: the border, scrollbars, and mouse hit-testing are unchanged.

## Tests

- Presenter geometry test: the content interior shifts exactly one column right when padded, the width narrows by one, and the tree column is never touched.
- Controller test: markdown and diff set the flag, syntax does not, and cycling a diff to the syntax view drops the gap.
- Full suite green (815 tests); `cargo fmt`/`clippy`/`audit` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@ All notable changes to this project are documented here. The format is based on
 - **In line-select mode, a mouse press now places the selection caret** (previously a click placed
   the line marker, and a double-click copied — copying is now always an explicit `Enter`/`y`).
 
+### Fixed
+- **Left gap in the rendered-markdown and diff views.** These views now sit one column in from the
+  content pane's left border instead of hugging it, matching the gap syntax-highlighted files
+  already get from their line-number gutter. Purely visual: the border, scrollbars, and mouse
+  hit-testing are unchanged.
+
 ## [1.10.0] - 2026-07-07
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,8 @@ All notable changes to this project are documented here. The format is based on
 ### Fixed
 - **Left gap in the rendered-markdown and diff views.** These views now sit one column in from the
   content pane's left border instead of hugging it, matching the gap syntax-highlighted files
-  already get from their line-number gutter. Purely visual: the border, scrollbars, and mouse
-  hit-testing are unchanged.
+  already get from their line-number gutter. Purely visual: the content interior is one column
+  narrower, and its text and mouse hit-testing shift together so clicks still map correctly.
 
 ## [1.10.0] - 2026-07-07
 
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -972,6 +972,17 @@ impl Controller {
         // mis-sizing the thumb / hiding the bar). Computed with the wrap we already have — no extra
         // tree walk.
         let content_rows = self.rendered_line_count_for(wrap);
+        // Inset the transformed views (rendered markdown / diff) one column from the left border so
+        // their delegate output — which starts at column 0 — doesn't hug it; syntax/plain files
+        // already get that gap from bat's line-number gutter, so they stay flush (no double gap).
+        // Keyed off the DISPLAYED content's file (`content_path`, the title's source of truth), so
+        // the gap switches in lockstep with the body — never off a still-loading selection.
+        let content_pad_left = self.content_path.as_ref().is_some_and(|p| {
+            matches!(
+                self.effective_mode(p),
+                ViewMode::RenderedMarkdown | ViewMode::Diff | ViewMode::FullDiff
+            )
+        });
         // Gutter width for a character selection's highlight (0 when not applicable); computed once
         // here so the line-select snapshot below stays a pure read.
         let sel_gutter = self.selection_gutter_len();
@@ -990,6 +1001,7 @@ impl Controller {
             tree_hscroll: self.tree_hscroll,
             content_rows,
             wrap,
+            content_pad_left,
             split_pct: self.split_pct,
             zoomed: self.zoomed,
             update_banner: self.update_banner(),
--- a/src/presenter.rs
+++ b/src/presenter.rs
@@ -68,6 +68,12 @@ pub struct ViewState {
     /// Wrap long content lines (prose: markdown / plain text) instead of truncating them.
     /// Off for diffs and code, whose column alignment must be preserved.
     pub wrap: bool,
+    /// Inset the content text one column from the left border. Set for the transformed views
+    /// (rendered markdown, diff) whose delegate output starts at column 0 and otherwise hugs the
+    /// border; syntax/plain files get the same visual gap for free from bat's line-number gutter,
+    /// so it stays off for them (no double gap). Applied identically by [`draw_content`] and
+    /// [`geometry`] so the drawn text rect and the hit-test geometry inset the border in lockstep.
+    pub content_pad_left: bool,
     /// The tree column's share of the width, as a percentage (the content pane takes the
     /// rest). Adjustable from the keyboard; used only in the wide two-column layout.
     pub split_pct: u16,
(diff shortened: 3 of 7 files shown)


smarzban:
## Empanel — Round 1

Findings: 4 total — 0 critical · 0 high · 3 medium · 1 low · 0 info.
_Reviewed by:_ holistic + lens-tests · models: deepseek:deepseek-v4-pro, ollama:glm-5.2, openai-codex:gpt-5.5, anthropic:claude-opus-4-8
⚠️ **Reviewer warnings:** holistic/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns))

### Must fix
- 🟠 **[MEDIUM]** \[test-coverage\] Mode-to-padding tests miss the loading/displayed-content transition — `tests/controller.rs:968` · 2/4 models · _by:_ lens-tests _(test-coverage)_
  The controller test only asserts \`content\_pad\_left\` after \`await\_marker\`, so it would not catch a refactor that keys the flag off the newly selected/effective mode before the new render lands. The risky behavior called out by the change is that the gap must track the displayed \`content\_path\` in lockstep with the body during async render transitions.
  _Fix:_ Use a blocking/slow \`ContentProvider\` to start on a rendered file, select or cycle to a different mode, assert the flag does not switch while the old body/loading state is still displayed, then release/poll and assert it changes only when the new render is applied.
- 🟠 **[MEDIUM]** \[test-coverage\] No rendering/snapshot test for content\_pad\_left = true case — `tests/presenter.rs:1179` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ lens-tests _(test-coverage)_
  The \`content\_pad\_left\_insets\_the\_transformed\_views\_one\_column\` test only asserts on \`geometry()\` output (the hit-test rect), never on the actual rendered buffer. The \`wide\_layout\_snapshot\` test (line \~1178) uses \`sample\_state()\` which has \`content\_pad\_left: false\`. While \`draw\_content\` and \`geometry\` both call \`content\_block()\`, there is no test that verifies they remain in lockstep — if someone accidentally reverted \`draw\_content\` to use \`Block::bordered()\` directly, the rendering would silently break but the geometry test would still pass. A snapshot test with the flag set to \`true\` would lock in the visual output.
  _Fix:_ Add a snapshot test (e.g., \`insta::assert\_snapshot!\`) that renders with \`content\_pad\_left: true\` and captures the buffer, or add a cell-level assertion in the existing geometry test that renders the buffer and verifies the content text starts one column further right.
  _Possible sibling (shared topic):_ \[correctness\] Block padding also changes the content hit area and horizontal bar — `src/presenter.rs:762`
  _Possible sibling (co-located — verify same defect):_ \[test-coverage\] Padded geometry is tested, but padded drawing is not — `tests/presenter.rs:1183`
- 🟠 **[MEDIUM]** \[test-coverage\] Padded geometry is tested, but padded drawing is not — `tests/presenter.rs:1183` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ lens-tests _(test-coverage)_
  The new presenter test only inspects \`geometry(...).content\_inner\`; it never renders with \`content\_pad\_left = true\` and verifies the first content cell/snapshot shifts. A regression that drops \`content\_block(state)\` from \`draw\_content\` while leaving \`geometry\` padded would still pass, even though the visual gap is the user-facing behavior.
  _Fix:_ Add a render-buffer or insta snapshot case with \`content\_pad\_left = true\` and compare the content text x-position against the unpadded case, ideally also asserting geometry's x matches the rendered first content column.
  _Possible sibling (co-located — verify same defect):_ \[test-coverage\] No rendering/snapshot test for content\_pad\_left = true case — `tests/presenter.rs:1179`

### Advisory (non-blocking)
- ⚪ **[LOW]** \[correctness\] Block padding also changes the content hit area and horizontal bar — `src/presenter.rs:762` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(correctness)_
  Applying the left gap as \`Block\` padding changes the block inner rect used for \`content\_bars\` and \`geometry\`, so the content viewport shrinks and the horizontal scrollbar/hit-test rect move with the text rather than leaving scrollbars and pane hit-testing unchanged.
  _Fix:_ Keep the content block's original inner rect for scrollbar layout and pane-level hit-testing, and apply the one-column inset only to the text/selection coordinate mapping; alternatively update the stated behavior/tests if the narrower viewport and shifted hit area are intentional.
  _Possible sibling (shared topic):_ \[test-coverage\] No rendering/snapshot test for content\_pad\_left = true case — `tests/presenter.rs:1179`

smarzban:

**Panel:** 4 holistic seats (deepseek-v4-pro, opus-4-8, glm-5.2, gpt-5.5) + `lens-tests` (gpt-5.5, deepseek-v4-pro) + deterministic scan. Holistic gpt/glm needed a re-run (transient Codex provider error; glm hit its turn cap — re-ran with a higher budget). Final panel is not thin: 4 holistic votes + 2 lens votes. Scan clean.

**Lens decision:** fired `lens-tests` (behavior changed, `suggest` flagged it). Skipped security/privacy/contracts/migrations/subtle-correctness/simplify/performance/infra/spec — this is a self-contained UI-layout change: no subprocess/untrusted-input surface, no data handling, no schema/DB/infra, and per-frame cost is a hashmap lookup (no perf concern). No spec doc exists for this small fix.

**What this round found** — all test-coverage, no correctness/security bugs:
- **Agree, fixing:** the geometry test asserts `geometry().content_inner` but never the *drawn buffer*, so a regression that dropped `content_block` from `draw_content` alone (leaving `geometry` padded) would pass. Adding a render-buffer assertion.
- **Agree, fixing:** the flag keys off the displayed `content_path`; the pre-render `None` state and the async displayed-vs-selected transition weren't tested. Adding a controller assertion + a slow-render transition test.
- **Dismissed (LOW, advisory):** "padding also shifts the hit area / hbar." Not a bug — `draw_content` and `geometry` both derive the interior from the *same* `content_block(state)`, so the text and hit-test rects shift in lockstep and clicks still map correctly (the reviewer's alternative would reintroduce an off-by-one). The one-column-narrower viewport is the intended cost. I will tighten the CHANGELOG line that overclaimed "hit-testing unchanged."

**Decision: 🔴 Request changes** — fixing the three test-coverage gaps, then Round 2 will verify.

smarzban:
## Empanel — Round 2

_Reviewed by:_ verify · models: deepseek:deepseek-v4-pro

### Progress since Round 1
✅ Resolved (3): \[test-coverage\] Mode-to-padding tests miss the loading/displayed-content transition; \[test-coverage\] No rendering/snapshot test for content\_pad\_left = true case; \[test-coverage\] Padded geometry is tested, but padded drawing is not — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

**What this PR does:** insets the rendered-markdown (`glow`) and diff (`delta`) content-pane views one column from the left border via a `content_pad_left` flag on `ViewState`, keyed off the *displayed* content's effective mode. Syntax/plain files stay flush (they already have `bat`'s gutter gap). The pad flows through a shared `content_block()` helper used by both `draw_content` and `geometry`, so the drawn text and the mouse hit-test rect shift in lockstep. Verified end-to-end in a live herdr pane against real `glow`/`delta`/`bat`.

**Fixed (cumulative across rounds):**
- Round 1 (verify): added a render-buffer test asserting the drawn gap (catches dropping `content_block` from `draw_content` alone), a `content_path=None` pre-render assertion, and a slow-render async-transition test (the gap must not flip ahead of the body). Tightened the CHANGELOG wording that overclaimed "hit-testing unchanged."

**Dismissed (advisory, code-checked):** the LOW "padding also shifts hit area / hbar" — not a bug: text and hit-test rects shift together via the shared `content_block`, so clicks map correctly; the one-column-narrower viewport is the intended cost.

**Deferred / not covered:** the bat-absent plain-text fallback for a code file has no gutter and no pad, so it would hug the border — an intentionally out-of-scope edge (owner confirmed "leave it as it is"); the original complaint was markdown/diff only.

**Coverage:** Round 1 ran the full 4-seat holistic panel (deepseek, opus, glm, gpt — the last two re-run after transient failures) + `lens-tests` (2 models) + deterministic scan (clean). Round 2 was single-model verification (deepseek). 817 tests green; fmt/clippy/audit clean.

**Decision: ✅ Approve** — verdict `pass` on HEAD `7b30a0a`.

PR #83, opened 2026-07-09
fix: open-in-tab no longer switches to a viewer in another workspace

smarzban:
## The bug

`prefix+shift+f` (the `open-file-viewer-tab` action) is idempotent by design: if a file-viewer tab is already open, it switches to it rather than duplicating. But the switch logic (`launch_decision_tab`) matched a `Files` viewer in **any** tab, including one in a *different workspace* — so invoking it from workspace B emitted `SWITCHTAB <workspace-A-tab>` and `herdr tab focus` yanked the user across to workspace A instead of opening a viewer in B.

## The fix

Scope the `SWITCHTAB` search to the **focused pane's own workspace** (herdr's explicit `workspace_id`, falling back to the `workspace:...` id prefix). A viewer that lives only in another workspace is left alone and a fresh viewer opens in the current workspace. If the focused workspace can't be determined, it degrades to `OPEN` — never a cross-workspace jump. The within-workspace open-or-switch-or-toggle idempotency is unchanged.

## Verification

- Reproduced live against a real `herdr pane list` (viewer in `w19`, focused in `wQ`): pre-fix would emit `SWITCHTAB w19:tB`; post-fix emits `OPEN`.
- Two new regression tests: viewer only in another workspace → `OPEN`; viewer in both this and another workspace → switch to the one in this workspace.
- Full suite green; `cargo fmt --check` + `cargo clippy --all-targets -D warnings` clean.

## Files

- `src/launch.rs` — workspace-scoped switch + `workspace_of` helper + 2 tests
- `CHANGELOG.md` — `Fixed` entry under `[Unreleased]`
- `docs/usage.md` — tab-launcher idempotency now described as workspace-scoped

Read-only; no new dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@ All notable changes to this project are documented here. The format is based on
   the line marker, and a double-click copied — copying is now always an explicit `Enter`/`y`).
 
 ### Fixed
+- **Open-in-tab (`prefix+shift+f`) no longer jumps to a viewer in another workspace.** When a file
+  viewer tab is already open in one workspace, invoking the tab action from a *different* workspace
+  now opens a fresh viewer in the current workspace instead of switching you to the other
+  workspace's tab. The open-or-switch idempotency stays intact within a workspace (still switches
+  to an existing viewer tab of the same workspace, and toggles off when you're on it).
 - **Left gap in the rendered-markdown and diff views.** These views now sit one column in from the
   content pane's left border instead of hugging it, matching the gap syntax-highlighted files
   already get from their line-number gutter. Purely visual: the content interior is one column
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -47,12 +47,16 @@ viewer via the same idempotent launcher. (Alternatively, `command` may invoke
 
 **Open in a tab instead of a split.** A second action, `open-file-viewer-tab`, opens the viewer
 in its **own tab** (`scripts/open-file-viewer-tab.sh`, `--placement tab`). Its launcher is
-idempotent *across tabs* — *open-or-switch-or-toggle*:
+idempotent *across the tabs of the current workspace* — *open-or-switch-or-toggle*:
 
-- no viewer anywhere → open it in a new tab (focused)
-- a viewer in another tab → **switch to that tab** (never a duplicate)
+- no viewer in this workspace → open it in a new tab (focused)
+- a viewer in another tab of this workspace → **switch to that tab** (never a duplicate)
 - a viewer in the current tab, not focused → focus it in place
 - the viewer already focused → close it (herdr auto-closes the emptied tab)
+
+The idempotency is scoped to the **current workspace**: a viewer already open in a *different*
+workspace is left where it is, and a fresh one opens here — the action reaches this workspace's
+viewer, it never pulls you across workspaces.
 
 Bind it to its own key — e.g. `prefix+shift+f` alongside `prefix+f` for the split:
 
--- a/scripts/open-file-viewer-tab.ps1
+++ b/scripts/open-file-viewer-tab.ps1
@@ -2,13 +2,16 @@
 #
 # Idempotent launcher for the file viewer in its own TAB -- used by the `open-file-viewer-tab-windows`
 # action and a herdr keybinding (e.g. `prefix+shift+f`). "Open-or-switch, toggle on repeat",
-# scoped across tabs:
-#   - no Files pane anywhere                  -> open the viewer in a new tab (focused)
-#   - a Files pane lives in another tab       -> switch to that tab (no duplicate viewer)
+# scoped across the tabs of the CURRENT WORKSPACE:
+#   - no Files pane in this workspace         -> open the viewer in a new tab (focused)
+#   - a Files pane in another tab of this
+#     workspace                               -> switch to that tab (no duplicate viewer)
 #   - a Files pane is in the current tab,
 #     but not focused                         -> focus it in place
 #   - the focused pane IS the Files pane      -> close it ("toggle off"; herdr auto-closes the
 #                                               now-empty tab -- verified)
+# A viewer open in a DIFFERENT workspace is left alone and a fresh one opens here (never switches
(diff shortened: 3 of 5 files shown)


smarzban:
## Empanel — Round 1

Findings: 3 total — 0 critical · 0 high · 1 medium · 2 low · 0 info.
_Reviewed by:_ holistic + lens-security · models: deepseek:deepseek-v4-pro, ollama:glm-5.2, openai-codex:gpt-5.5, anthropic:claude-opus-4-8
⚠️ **Reviewer warnings:** holistic/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns))

### Must fix
- 🟠 **[MEDIUM]** \[security\] Workspace scoping trusts inconsistent workspace\_id — `src/launch.rs:126` · 2/4 models · _by:_ holistic, lens-security _(security)_
  A crafted pane-list can set the focused pane's or viewer pane's \`workspace\_id\` to the current workspace while the viewer's \`tab\_id\` still names another workspace (for example \`w19:tB\`). The comparison then passes and emits \`SWITCHTAB w19:tB\`; \`is\_flag\_safe\` only prevents option injection and does not prevent this cross-workspace focus.
  _Fix:_ Derive/validate the workspace from the action target's \`tab\_id\` prefix, and reject panes where \`workspace\_id\`, \`tab\_id\`, and \`pane\_id\` disagree or where the id has no valid workspace prefix before emitting \`SWITCHTAB\`.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[docs\] Windows tab launcher header still describes cross-workspace idempotency — `scripts/open-file-viewer-tab.ps1:5` · 1/4 models · _by:_ holistic _(maintainability)_
  The PowerShell launcher uses the same \`--launch-decision-tab\` logic, but its header still documents global across-tabs switching rather than current-workspace scoping. That leaves the Windows launcher documentation inconsistent with the actual behavior and the updated usage docs.
  _Fix:_ Mirror the updated wording from \`docs/usage.md\`: no viewer in this workspace opens a tab, same-workspace viewers are switched to, and other-workspace viewers are left alone.
- ⚪ **[LOW]** \[documentation\] Stale comment in open-file-viewer-tab.sh header doesn't reflect new workspace scoping — `scripts/open-file-viewer-tab.sh:5` · 4/4 models · _by:_ holistic _(documentation)_
  The script's header comment still says "no Files pane anywhere" and "a Files pane lives in another tab" without mentioning workspace scoping. The actual behavior is correct (the decision comes from the Rust binary via --launch-decision-tab), but anyone reading the script to understand the semantics will be misled. The docs/usage.md was updated correctly.
  _Fix:_ Update the header comment to reflect workspace scoping, e.g.: "no Files pane in this workspace → open a new tab" and "a Files pane lives in another tab of this workspace → switch to that tab".

smarzban:

I ran the full holistic panel (deepseek-v4-pro, glm-5.2, gpt-5.5, opus-4-8) plus `lens-security` (gpt-5.5, deepseek-v4-pro) — fired because the change parses the host's untrusted `pane list` JSON and feeds an id into a `herdr tab focus <id>` argv — and the deterministic scan (clean; secrets tier ran).

**What I agree with and am fixing:**

- **[medium/security] Workspace scoping trusted an inconsistent `workspace_id`.** `workspace_of` read the separate `workspace_id` field, but the launcher switches on `tab_id` — so a `pane list` where those disagree (`workspace_id:"wQ"`, `tab_id:"w19:tB"`) could pass the scope check yet emit `SWITCHTAB w19:tB`. The codebase already treats host ids as untrusted (that is what `is_flag_safe` is for), so the fix derives the scope from the *same id we act on* (`tab_id` prefix, via `split_once`), and drops the unused `workspace_id` field. This also resolves the low "unqualified id treated as a workspace" nit (no `:` → unknown → `OPEN`).
- **[low/docs, 4/4 models] Stale header comments** in `scripts/open-file-viewer-tab.sh` and (`1/4`) its Windows `.ps1` twin still described unscoped across-tabs switching. Both headers updated to describe current-workspace scoping.
- **[low/test-coverage] Derivation path untested.** Added two regression tests: a spoofed `workspace_id` cannot force a cross-workspace switch, and an unqualified id opens rather than guessing.

**Dismissed:** none.

**Decision: 🔴 Request changes** — matches the Round 1 `block` verdict. Fixing all of the above, then re-verifying in Round 2.

smarzban:
## Empanel — Round 2

_Reviewed by:_ verify · models: deepseek:deepseek-v4-pro

### Progress since Round 1
✅ Resolved (1): \[security\] Workspace scoping trusts inconsistent workspace\_id — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

Round 2 was a single-model verification pass (deepseek-v4-pro) scoped to Round 1's findings + `git diff 99494af...HEAD`, plus a fresh deterministic scan. It returned **0 findings** — every Round 1 item is resolved at HEAD and no fix introduced a regression. Scan clean.

**What this PR implements:** `prefix+shift+f` (the `open-file-viewer-tab` action) no longer switches to a file-viewer tab that lives in a *different* workspace. The switch is scoped to the focused pane's own workspace; a viewer that exists only in another workspace is left alone and a fresh one opens in the current workspace. Within-workspace open-or-switch-or-toggle idempotency is unchanged.

**Fixed (cumulative across rounds):**
- Cross-workspace jump on the tab launcher (the reported bug).
- Round 1: scope now derived from the acted-on `tab_id` (self-consistent, robust to an inconsistent `workspace_id`); unqualified ids treated as unknown → `OPEN`; launcher `.sh`/`.ps1` header comments refreshed; two regression tests added.

**Verification:** live-checked against a real `herdr pane list` with `Files` viewers open in three workspaces — the launcher correctly switches to the viewer in the current workspace and ignores the other two. Full test suite green; `cargo fmt --check` + `clippy -D warnings` clean.

**Deferred / not covered:** none. Read-only change, no new dependencies.

**Decision: ✅ Approve** — matches the Round 2 `pass` verdict.

PR #87, opened 2026-07-10
fix: render wide markdown tables to fit the pane, with a `w` wide/scroll view

smarzban:
## What

Wide tables in the rendered-markdown view used to shatter: `glow` was invoked with `-w 0` (wrapping disabled), so a table was laid out at its natural (very wide) width and the content pane's line-wrapping then broke it apart across the border rows. This PR makes tables render correctly, and adds a way to read cells that don't fit.

Two commits:

1. **fit-to-pane (default).** Point glow's `-w` at the content pane's drawable width, so glow sizes the columns and wraps prose to fit (borders intact; over-long cells ellipsized with `…`). Because glow's output is exactly the pane width, the Presenter's re-wrap becomes a no-op. A resize or split-bar drag reflows the table to the new width, preserving scroll position and any active search.

2. **`w` = wide, horizontally-scrollable view.** Fitting truncates cells, so the hidden text was unreachable. `w` now toggles rendered markdown between the fit view and a wide view that renders the table at full natural width, where the pane scrolls sideways (`←`/`→` or the scrollbar) to reveal every cell. Previously `w` did nothing for markdown. It stays a uniform wrap on/off preference, so unwrapping a table never springs an unexpected wrap onto a code file viewed next.

## Why not "fit everything by wrapping cells"

glow cannot wrap table cells (it only ellipsizes), and no renderer we delegate to does. Fitting all content and reflowing would mean rendering tables ourselves, which cuts against the delegate-rendering principle (ADR-0001). The fit view + the `w` wide/scroll escape hatch gives both "adapts to the pane" and "read everything", with zero custom rendering.

## How it works

- `ContentProvider` gains a defaulted `render_at_width`; the render worker calls it with the pane width carried on `RenderJob`. Only the live `LiveContent` overrides it (threads the width into glow's `-w`), so the test doubles are untouched.
- `wrap_override` becomes `Option<bool>` (None = per-mode default, Some(true/false) = force on/off). `md_wrap_width` hands glow the pane width only in the fit view; in the wide view glow keeps its natural-width `-w 0` output and the existing content h-scroll takes over.
- A width change (resize) or a `w` toggle re-renders markdown through one scroll/search-preserving reflow path, distinct from the selection-change render that resets view state.

## Scope / safety

- Read-only; no new dependencies.
- Rendering logic only; no UI-string or manifest changes.

## Tests

- Width threading: hermetic (shell renderer echoing `-w`) plus a real-glow check that no rendered line exceeds the wrap width (the no-gap property).
- Resize reflow vs. height-only no-op vs. non-markdown no-op; scroll + committed search preserved across a reflow.
- `w` flips fit <-> wide (both `view_state().wrap` and the width handed to glow); unwrapping markdown does not force-wrap a code file viewed next.

Full suite green; fmt/clippy/audit clean. Verified live in herdr (fit view, `w` to wide, scroll right to read the full Notes column, `w` back to fit).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,20 +6,15 @@ All notable changes to this project are documented here. The format is based on
 
 ## [Unreleased]
 
-### Changed
-- **`w` now toggles rendered markdown between fit-to-pane and a wide, horizontally-scrollable
-  view.** In the wide view a table renders at its full natural width with every cell intact, and
-  the content pane scrolls sideways (`←`/`→` or the scrollbar) to reveal it; press `w` again to
-  return to the fit view. Previously `w` had no effect on markdown (which always wrapped). `w`
-  remains a uniform wrap on/off preference, so it never springs an unexpected wrap on a code file.
-
 ### Fixed
-- **Wide tables in the rendered-markdown view.** In the default (fit) view a table is now laid out
-  to fit the content pane (columns sized to the pane width, over-long cells ellipsized with `…`)
-  instead of overflowing at its natural width and being shattered across the border rows by the
-  pane's line-wrapping. Resizing the pane (or dragging the split bar) reflows the table to the new
-  width, keeping your scroll position and any active search. To read a truncated cell in full,
-  press `w` for the wide, horizontally-scrollable view (see above).
+- **Wide tables in the rendered-markdown view no longer shatter.** A table is now laid out to fit
+  the content pane (columns sized to the pane width, over-long cells ellipsized with `…`) instead
+  of overflowing at its natural width and being broken across the border rows by the pane's
+  line-wrapping. Resizing the pane (or dragging the split bar) reflows the table, preserving your
+  scroll position and any active search. Press `w` to switch to a wide, horizontally-scrollable
+  view that renders the table at full natural width with every cell intact (`←`/`→` or the
+  scrollbar to pan); press `w` again to return. (`w` remains a uniform wrap on/off preference, so
+  it never springs an unexpected wrap on a code file.)
 
 ## [1.11.0] - 2026-07-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,20 @@ All notable changes to this project are documented here. The format is based on
 
 ## [Unreleased]
 
+### Changed
+- **`w` now toggles rendered markdown between fit-to-pane and a wide, horizontally-scrollable
+  view.** In the wide view a table renders at its full natural width with every cell intact, and
+  the content pane scrolls sideways (`←`/`→` or the scrollbar) to reveal it; press `w` again to
+  return to the fit view. Previously `w` had no effect on markdown (which always wrapped). `w`
+  remains a uniform wrap on/off preference, so it never springs an unexpected wrap on a code file.
+
 ### Fixed
-- **Wide tables in the rendered-markdown view.** A table is now laid out to fit the content pane
-  (columns sized to the pane width, over-long cells ellipsized with `…`) instead of overflowing at
-  its natural width and being shattered across the border rows by the pane's line-wrapping.
-  Resizing the pane (or dragging the split bar) reflows the table to the new width, keeping your
-  scroll position and any active search.
+- **Wide tables in the rendered-markdown view.** In the default (fit) view a table is now laid out
+  to fit the content pane (columns sized to the pane width, over-long cells ellipsized with `…`)
+  instead of overflowing at its natural width and being shattered across the border rows by the
+  pane's line-wrapping. Resizing the pane (or dragging the split bar) reflows the table to the new
+  width, keeping your scroll position and any active search. To read a truncated cell in full,
+  press `w` for the wide, horizontally-scrollable view (see above).
 
 ## [1.11.0] - 2026-07-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project are documented here. The format is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **Wide tables in the rendered-markdown view.** A table is now laid out to fit the content pane
+  (columns sized to the pane width, over-long cells ellipsized with `…`) instead of overflowing at
+  its natural width and being shattered across the border rows by the pane's line-wrapping.
+  Resizing the pane (or dragging the split bar) reflows the table to the new width, keeping your
+  scroll position and any active search.
+
 ## [1.11.0] - 2026-07-09
 
 Mouse-driven text selection in the content pane, plus content copy in line-select mode.
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ PowerShell launcher scripts.
 | `Y` | Copy the selected file's **absolute** path to the clipboard |
 | `Tab` | Move focus between the tree and content columns |
 | `<` / `>` | Narrow / widen the tree column (move the divider) |
-| `w` | Toggle line wrapping for the content pane |
+| `w` | Toggle line wrapping for the content pane. For rendered markdown this switches between the fit-to-pane view (wide tables sized to fit, over-long cells shown as `…`) and a wide view that renders tables at full width and scrolls horizontally (`←`/`→`) so you can read every cell |
 | `z` | Zoom: hide the tree so the content pane fills the frame; press again (or `q`/`Esc`) to restore the two-column layout |
 | `r` | Refresh git state: pick up changes made outside the viewer (a merge / pull / commit elsewhere) |
 | `W` (Shift+`w`) | **Switch worktree**: open a picker of the repo's git worktrees and re-root the viewer to the one you pick (read-only; no branch checkout). Marks the current worktree and pre-selects the one with an active herdr agent; `↑`/`↓` move, `←`/`→` scroll long paths, `Enter` switches, `Esc` cancels |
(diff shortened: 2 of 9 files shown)


smarzban:
## Empanel — Round 1

Findings: 5 total — 0 critical · 0 high · 0 medium · 4 low · 1 info.
_Reviewed by:_ holistic + lens-security + lens-subtle-correctness + lens-tests · models: deepseek:deepseek-v4-pro, ollama:glm-5.2, anthropic:claude-opus-4-8
⚠️ **Coverage:** 7/9 planned reviewer passes voted — lost: holistic/pi:openai-codex:gpt-5.5 (holistic: provider error: Your authentication token has been invalidated. Please try signing in again.); lens-security/pi:openai-codex:gpt-5.5 (lens-security: provider error: Your authentication token has been invalidated. Please try signing in again.)
⚠️ **Reviewer warnings:** holistic/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns); near-cap: submitted after the near-cap nudge — the review hit its turn budget and may be incomplete)

### Advisory (non-blocking)
- ⚪ **[LOW]** \[test-coverage\] recompute\_committed\_search clamping on reduced match count untested — `src/controller/infile.rs:303` · 2/3 models · _by:_ holistic, lens-tests _(test-coverage)_
  \`recompute\_committed\_search\` clamps \`current\` with \`prev.current.min(matches.len().saturating\_sub(1))\` — when a width reflow reduces the match count (lines merge/split differently), the ordinal is clamped. The integration test \`a\_width\_reflow\_preserves\_scroll\_and\_recomputes\_a\_committed\_search\` uses a \`WidthProbe\` that emits identical line tokens, so the match count never shrinks. The clamping path is untested.
  _Fix:_ Modify the test (or add a variant) where the reflowed body has fewer matching lines than the original, and verify \`current\` is correctly clamped rather than indexing out of bounds.
- ⚪ **[LOW]** \[test-coverage\] toggle\_wrap on non-markdown file still calls rerender\_markdown\_reflow — no test guards against regressions in the no-op path — `src/controller/mod.rs:1786` · 1/3 models · _by:_ lens-tests _(test-coverage)_
  \`toggle\_wrap\` unconditionally calls \`rerender\_markdown\_reflow()\`. For a code file (SyntaxContent) the inner guard returns early, making this a no-op. No test verifies that toggling \`w\` on a code file doesn't accidentally dispatch a render (e.g., if someone removes the mode-guard in \`rerender\_markdown\_reflow\`). The existing \`w\_toggles\_markdown\_between\_the\_fit\_and\_the\_wide\_unwrapped\_view\` test only covers markdown.
  _Fix:_ Add a test: select a \`.rs\` file, toggle wrap, and verify no extra render jobs were dispatched (e.g., use a WidthProbe and assert the recorded widths vector is unchanged after the toggle).
- ⚪ **[LOW]** \[performance\] Resize path now does an extra gitignore-aware tree walk per frame via effective\_wrap() — `src/controller/mod.rs:975` · 1/3 models · _by:_ holistic _(performance)_
  set\_content\_viewport now calls self.effective\_wrap() on every viewport change, and effective\_wrap() -\> wrap\_for(self.tree.selected()) re-runs the gitignore-aware filesystem walk (the same walk view\_state deliberately avoids doubling). When the reflow actually fires, rerender\_markdown\_reflow calls self.tree.selected() again — so an interactive terminal resize or split-bar drag (viewport changes every frame) adds 1-2 full tree walks per frame on top of view\_state's visible\_nodes() walk. On a large repository this is new per-frame work that can make resize/drag janky where it previously was not.
  _Fix:_ Avoid tree.selected() in the resize hot path: cache the selected node's effective mode/wrap from the already-computed visible\_nodes() in view\_state (or store the last selected Node), and have set\_content\_viewport/effective\_wrap read that cache instead of re-walking. At minimum, rerender\_markdown\_reflow could reuse the node effective\_wrap already resolved rather than calling tree.selected() a second time.
- ⚪ **[LOW]** \[test-coverage\] with\_wrap\_width trailing -w edge case untested — `src/render.rs:185` · 1/3 models · _by:_ lens-tests _(test-coverage)_
  \`with\_wrap\_width\` handles the case where \`-w\` is the last argument (trailing, no value following) by pushing the width value after it. This edge case (\`out.get\_mut(i + 1)\` is \`None\`) has no test. If the logic were refactored incorrectly, a regression wouldn't be caught.
  _Fix:_ Add a test: \`with\_wrap\_width(&\["glow", "-w"\], 70)\` should produce \`\["glow", "-w", "70"\]\`.
- ⚪ **[INFO]** \[test-coverage\] ContentProvider::render\_at\_width default impl never directly verified — `src/controller/mod.rs:123` · 1/3 models · _by:_ lens-tests _(test-coverage)_
  The default trait implementation of \`render\_at\_width\` ignores \`width\` and delegates to \`render\`. The worker thread now calls \`render\_at\_width\` instead of \`render\`, so all non-markdown test stubs depend on the default impl forwarding correctly. No test explicitly asserts that \`render\_at\_width(Some(42), ...)\` on a stub returns the same output as \`render(...)\`. Though existing tests would break if this were wrong, a dedicated test would make the contract explicit.
  _Fix:_ Add a unit test on a stub \`ContentProvider\` that asserts \`render\_at\_width(p, m, r, Some(42))\` produces the same \`RenderResult\` as \`render(p, m, r)\`.

smarzban:

**Decision: ✅ Approve** (agrees with the deterministic `pass` verdict.)

### What this PR implements
Fixes wide markdown tables shattering in the content pane by pointing `glow`'s `-w` at the pane's drawable width (fit-to-pane, reflows on resize, over-long cells ellipsized), and adds `w` as an escape hatch that switches rendered markdown to a full-width, horizontally-scrollable view so every cell is readable. Read-only, no new dependencies.

### Verification
Deterministic scan clean. Full panel ran; I read the code behind every finding. All 5 findings are LOW/info and non-gating. I confirmed in the code:
- **Empty-matches after a reflow is safe.** `recompute_committed_search` can produce `matches = []` with `current = 0`, but `next_match`/`prev_match` (`src/controller/infile.rs:221`/`245`) guard with `!s.matches.is_empty()` before indexing, so there is no out-of-bounds path. The corresponding finding is a test-coverage gap, not a bug.
- **The perf finding's premise is inaccurate.** `tree.selected()` runs an in-memory `visible_nodes()` traversal, not a gitignore filesystem walk (that happens once at tree construction via the `ignore` crate); `effective_mode()` reads only in-memory maps. The real cost is a redundant in-memory traversal during an active resize, bounded by visible-node count and dominated by the `glow` re-render it gates. LOW/advisory.
- **The opus finding is a design tradeoff** (the reviewer concurs): a resize preserves the scroll position, so the current search match can be off-screen. Intended behavior.

### Deferred / not covered (advisory, non-blocking)
- Test-coverage gaps (all on already-verified paths): `recompute_committed_search` clamping with a *reduced* match count (2/3 models), `toggle_wrap` no-op on a non-markdown file, `with_wrap_width` trailing-`-w`, and a direct assertion of the `render_at_width` default impl. Cheap, pure-additive tests; happy to fold them in if wanted.
- Micro-optimization: avoid the second `tree.selected()` in the resize path.

### Coverage note
gpt-5.5 (OpenAI Codex) non-voted on both its seats (holistic + lens-security) — its OAuth token expired mid-run. The panel still carried 3 diverse holistic models (deepseek, glm-5.2, opus-4-8) plus the security lens re-run on opus, so it is not thin, but the security-sniper seat's shot was lost. Re-auth (`empanel auth login openai-codex`) would let that seat be re-run.

GitHub CI is green across ubuntu/macOS/windows + fmt/clippy + cargo-audit.

smarzban:
## Empanel — Round 2

Findings: 1 total — 0 critical · 0 high · 0 medium · 1 low · 0 info.
_Reviewed by:_ holistic + lens-security · models: openai-codex:gpt-5.5

### Progress since Round 1
✅ Resolved (0)
⏳ Still blocking (0)
🆕 New / regressed (0)

### Advisory (non-blocking)
- ⚪ **[LOW]** \[correctness\] Markdown wrap width can be too wide when a vertical scrollbar appears — `src/controller/mod.rs:1927` · 1/1 models · _by:_ holistic _(correctness)_
  The width handed to glow is the text width from the previous draw; when the newly rendered markdown is tall enough to need a vertical scrollbar, the presenter then steals two columns for the scrollbar/gap and wraps glow's just-padded lines in the interim frame. This can leave the exact table-shattering/blank-gap behavior visible until a second reflow completes, especially noticeable with a slow markdown renderer.
  _Fix:_ Account for the scrollbar gutter before dispatching the width-sensitive render (or immediately schedule/apply a corrective reflow before drawing when the rendered result will overflow vertically), so the first width-aware markdown result is rendered to the actual text rect width.

smarzban:
## Orchestrator review — Round 2

**Decision: ✅ Approve** (agrees with the deterministic `pass` verdict.)

### What changed since Round 1
- Backfilled the 4 LOW/info test-coverage gaps the panel flagged (test-only commit, no production change): `with_wrap_width` trailing-`-w`, `toggle_wrap` no-op on a code file, the `render_at_width` default-impl forwarding contract, and the 2-model one: a committed-search ordinal is clamped when a reflow shrinks the match count (driven by a width-dependent match-count provider that reaches the path the identical-token test could not).
- Re-ran gpt-5.5 (OAuth restored), closing the Round 1 coverage gap: it now voted on both its seats. Cumulative holistic coverage is 4 distinct models (deepseek, glm-5.2, opus-4-8, gpt-5.5); security lens clean on both opus and gpt-5.5.

### The one new finding (advisory)
gpt-5.5 raised a LOW correctness edge: the first width-aware markdown render can be 2 columns too wide when the content is tall enough to summon a vertical scrollbar (the presenter reserves the gutter after glow already padded to the pre-scrollbar width), so a tall table can briefly show the gap on its first frame. I confirmed in code that this **self-corrects**: once the scrollbar appears, `text.width` shrinks by 2, `set_content_viewport` sees the change, and the existing reflow re-renders glow to the exact width (the vbar decision is monotonic under wrap, so it converges, no oscillation). It is a transient flicker on first display of a tall table, not a persistent break, hence LOW/advisory. A tidy follow-up would reserve the gutter (or apply a corrective reflow) before the first width-aware dispatch.

### Cumulative status
Clean PASS across both rounds. No critical/high/medium; deterministic scan clean each round; GitHub CI green (ubuntu/macOS/windows + fmt/clippy + cargo-audit). All findings are advisory and either fixed (the test gaps) or verified non-blocking in code (perf premise inaccurate; opus + gpt edges are transient/by-design). Deferred follow-ups: the first-frame scrollbar-gutter reflow and the resize-path micro-optimization.

PR #89, opened 2026-07-11
feat: `Z` toggles a file full-screen (in-pane zoom + herdr pane zoom)

smarzban:
## Summary

`Z` (Shift+`z`) toggles a file "full-screen". The first press opens the selected file like `Enter`
(content fills the viewer, content pane focused) and zooms the viewer's herdr pane to fill the whole
terminal, so the file takes over the entire screen instead of just the split. Press `Z` again to
return to the normal two-column split.

## Why

Opening a file already fills the viewer's own pane, but the viewer lives in a herdr split, so a long
file still only gets half the terminal. `Z` drives herdr's own pane zoom so you can read a file
across the full screen without reaching for herdr's pane-zoom key, and toggle straight back.

## How it works

- New `Intent::OpenFullscreen`, bound to `Z` in the input map. Lowercase `z` is unchanged (in-pane
  zoom only); `Z` on a directory just expands/collapses like `Enter`.
- The toggle reads herdr's real pane-zoom state (`herdr pane layout --current` ->
  `result.layout.zoomed`) rather than a local flag, so it stays in sync even when you zoom/un-zoom
  the pane with herdr's own key (`prefix+z`). It then issues `herdr pane zoom --current --on|--off`
  through a new best-effort `HerdrCli::run` seam.
- The argv is entirely static (no interpolated pane id), so there is no option-injection surface.
  `--current` resolves to the focused pane, which is always the viewer while it is handling a
  keystroke.
- Read-only with respect to files and git: a pane zoom is a herdr layout op. It degrades to the
  in-pane zoom when herdr is absent or a call fails (nothing panics).
- Verified against the installed herdr 0.7.3 CLI (`pane zoom [--current] [--on|--off]`,
  `pane layout --current`).

## Tests

- Input mapping: `Z` maps to `OpenFullscreen`, lowercase `z` stays `ToggleZoom`, Ctrl+Z is inert.
- Intent completeness (closed-set + no-file-mutation invariants).
- Controller: a file zooms (queries state, then `--on`); a second press un-zooms (`--off`) and
  restores the split; a directory only activates; an absent herdr still zooms in-pane.

## Docs

README `## Keys` row + Shift-keys note, CHANGELOG (Unreleased), ARCHITECTURE herdr-module row. Also
adds an AGENTS.md rule to check herdr's live docs/CLI before scoping or building host-boundary work.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the test suite are green
locally (the only local reds are the file-I/O-heavy `index_perf` scaling tests, which flake on
tmpfs pressure during the full concurrent run and pass in isolation; untouched by this change).


smarzban:
## Empanel — Round 1
🚫 **BLOCK** — 4 blocking finding(s) must be resolved or justified.

Findings: 6 total — 0 critical · 0 high · 4 medium · 2 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-tests · models: deepseek:deepseek-v4-pro, ollama:glm-5.2, openai-codex:gpt-5.5, anthropic:claude-opus-4-8
⚠️ **Reviewer warnings:** holistic/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns); near-cap: submitted after the near-cap nudge — the review hit its turn budget and may be incomplete); lens-tests/ollama:glm-5.2 (budget-unmetered: run cost is not fully metered (provider reported no usable cost for one or more turns))

### Must fix
- 🟠 **[MEDIUM]** \[lifecycle\] Quit/exit leaves the host pane zoomed after the viewer exits — `src/controller/mod.rs:1477` · 3/4 models · _by:_ holistic _(correctness)_
  \`host\_zoom(true)\` is only ever reversed inside \`open\_fullscreen\`'s unzoom branch. There is no call to \`host\_zoom(false)\` on the quit/exit path (grep shows \`host\_zoom\` is referenced only at its definition and the two open\_fullscreen call sites). So pressing \`Z\` to full-screen then \`q\`/\`Esc\`/Ctrl-C to quit leaves herdr's pane zoomed after the viewer process is gone — a host-state leak at teardown. This is the new-resource-without-cleanup failure mode: the change creates host zoom state but no teardown releases it on exit.
  _Fix:_ Call \`self.host\_zoom(false)\` on the quit path (the Intent::Quit / run-loop exit) so the viewer always restores the host pane to the split before it tears down, best-effort like the other host calls.
- 🟠 **[MEDIUM]** \[concurrency\] \`host\_pane\_zoomed\`/\`host\_zoom\` spawn blocking subprocesses on the input thread with no timeout on every \`Z\` press — `src/controller/mod.rs:1498` · 3/4 models · _by:_ holistic, lens-tests _(concurrency)_
  \`host\_pane\_zoomed()\` calls \`herdr.run\_json(...)\` → \`Command::new(program).args(args).output()\` (a blocking, unbounded subprocess spawn) and \`host\_zoom()\` does the same — both run synchronously inside \`handle()\` on the input thread. Each \`Z\` press does TWO blocking spawns (read \`pane layout --current\`, then \`pane zoom\`). The architecture is explicit that input must never block (AC-22/AC-23); a slow, wedged, or absent-on-PATH herdr binary freezes the TUI indefinitely (no timeout on \`Command::output\`). Unlike the one-shot worktree-picker query, \`Z\` is a repeat-pressed toggle, so the blocking call is on a hot path.
  _Fix:_ Either move the host-zoom read/write off the input thread (queue like the render/status workers), or wrap the spawn in a bounded timeout (e.g. \`wait\_timeout\`/\`Command::output\` with a deadline) so a wedged herdr degrades to the in-plugin zoom within the responsiveness budget rather than freezing input.
- 🟠 **[MEDIUM]** \[lifecycle\] Close/Esc does not tear down the host pane zoom opened by Z — `src/controller/mod.rs:1762` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(correctness)_
  After \`Z\` succeeds, both \`self.zoomed\` and herdr's host pane zoom are on, but \`close\_or\_unzoom\` only clears the local zoom flag. Pressing \`q\`/Esc from fullscreen therefore restores the in-pane split while leaving the herdr pane zoomed, so the host layout remains fullscreen until the user discovers another way to unzoom it.
  _Fix:_ Track that \`OpenFullscreen\` enabled host zoom (or query it here) and best-effort issue \`pane zoom --current --off\` when backing out of local zoom; cover the \`Z\` then Esc/q path in controller tests.
  _Possible sibling (shared topic):_ \[test-coverage\] No test covers the desync/teardown cases the new host-zoom state introduces — `tests/controller.rs:1853`
- 🟠 **[MEDIUM]** \[lifecycle\] Re-root resets in-plugin zoom but never un-zooms the host pane — `src/controller/mod.rs:723` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(correctness)_
  re\_root sets \`self.zoomed = false\` (line 723) but does NOT call \`self.host\_zoom(false)\`. If the user entered full-screen via \`Z\` (host pane zoomed=true) and then switches worktree with \`W\`, the host pane stays zoomed while the plugin resets to the two-column split — a visible desync the user must manually undo. The change introduces new session state (host pane zoom) but the existing re-root teardown path was not updated to clean it up; \`herdr\` is explicitly session-level and survives re-root, so the zoom survives too while the plugin view does not.
  _Fix:_ In re\_root, alongside \`self.zoomed = false\`, call \`self.host\_zoom(false)\` to restore the host pane to the split before re-rooting (best-effort, mirroring open\_fullscreen's unzoom branch).
  _Possible sibling (shared topic):_ \[test-coverage\] \`HerdrCli::run\` default impl never tested through \`LiveHerdr\` — only exercised via stubs in controller tests — `src/herdr.rs:35`

### Advisory (non-blocking)
- ⚪ **[LOW]** \[test-coverage\] \`HerdrCli::run\` default impl never tested through \`LiveHerdr\` — only exercised via stubs in controller tests — `src/herdr.rs:35` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ lens-tests _(test-coverage)_
  The new \`run\` method uses a trait default \`self.run\_json(args).map(\|\_\| ())\`. \`tests/herdr.rs\` tests \`LiveHerdr::run\_json\` extensively (argv correctness, non-zero exit → Err), but never calls \`run\` on \`LiveHerdr\`. The controller tests exercise \`run\` only through the \`PaneZoomFake\` (a separate \`HerdrCli\` impl), so the production wiring — \`LiveHerdr::run\` → \`LiveHerdr::run\_json\` → \`CommandRunner::run\` → shell out to herdr — has no hermetic coverage. A future override of \`run\` on \`LiveHerdr\` (e.g., a separate optimized code path) could be wrong without any unit test noticing.
  _Fix:_ Add a test in \`tests/herdr.rs\` that calls \`cli.run(&\["pane", "zoom", "--current", "--on"\])\` on a \`RecordingRunner\`-backed \`LiveHerdr\` and asserts the recorded argv and success/error propagation.
  _Possible sibling (shared topic):_ \[lifecycle\] Re-root resets in-plugin zoom but never un-zooms the host pane — `src/controller/mod.rs:723`
- ⚪ **[LOW]** \[test-coverage\] No test covers the desync/teardown cases the new host-zoom state introduces — `tests/controller.rs:1853` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic, lens-tests _(test-coverage)_
  The added tests cover the happy paths (Z opens/zooms, Z again unzooms, Z on a directory, Z without herdr) but none exercise the interleaving or lifecycle cases where the new host-zoom state actually breaks: \`z\` then \`Z\` (or \`Z\` then \`z\`), a re-root while host-zoomed, or quit while host-zoomed. Those are exactly the transitions where the host pane and in-plugin view desync, so the regressions the review found are unguarded.
  _Fix:_ Add tests asserting host-zoom symmetry after \`z\`/\`Z\` interleaving, that a re-root issues \`pane zoom --current --off\`, and that the quit path restores the host pane.
  _Possible sibling (shared topic):_ \[lifecycle\] Close/Esc does not tear down the host pane zoom opened by Z — `src/controller/mod.rs:1762`

smarzban:
## Empanel — Round 2

Findings: 2 total — 0 critical · 0 high · 2 medium · 0 low · 0 info.
_Reviewed by:_ verify · models: openai-codex:gpt-5.5
✅ **Dismissal audit:** all 2 dismissal justification(s) held against the code.

### Progress since Round 1
✅ Resolved (4): \[lifecycle\] Quit/exit leaves the host pane zoomed after the viewer exits; \[concurrency\] \`host\_pane\_zoomed\`/\`host\_zoom\` spawn blocking subprocesses on the input thread with no timeout on every \`Z\` press; \[lifecycle\] Close/Esc does not tear down the host pane zoom opened by Z; \[lifecycle\] Re-root resets in-plugin zoom but never un-zooms the host pane — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (2): \[lifecycle\] Z can unzoom a pane the user had already zoomed outside the viewer; \[concurrency\] Z still runs an unbounded herdr subprocess on the input thread

- **[MEDIUM]** \[lifecycle\] Z can unzoom a pane the user had already zoomed outside the viewer — `src/controller/mod.rs:1544`
  _Dismissed:_ Single-model advisory medium; the behavior is a deliberate, coherent trade-off and the only real defect (a doc overclaim) is fixed in this commit. The flag-based toggle means pressing Z while the pane is already full-screen via herdr's own prefix+z makes the viewer ADOPT that full-screen (the --on is a no-op) and a later Z/Esc/z issues --off. That is correct for a toggle that must work with no live herdr: Z consistently toggles the full-screen state of the pane it lives in. The alternative (a pane layout query on every Z to tell the users zoom from ours) is the round-1 approach, which was heavier (two subprocess calls per press) and could not toggle back at all when herdr was absent (the round-1 gpt/opus finding this fix resolves). The host\_zoomed field doc previously overclaimed never un-zoom a pane the user zoomed themselves; that doc is corrected in commit d913907 to describe the adopt-on-Z behavior, so doc and code now agree. The edge only manifests when a user mixes herdr prefix+z with Z on the viewer pane, and the outcome (Z/Esc returns to the split) is reasonable. Not a merge-blocker.
- **[MEDIUM]** \[concurrency\] Z still runs an unbounded herdr subprocess on the input thread — `src/controller/mod.rs:1571`
  _Dismissed:_ Single-model advisory medium, and a pre-existing accepted pattern rather than a new risk. The synchronous herdr-CLI-on-the-input-thread call already ships in the worktree picker (src/controller/picker.rs:162-163 runs run\_json(\[worktree,list\]) and run\_json(\[agent,list\]) inline when W opens the picker). This PR makes Z issue AT MOST ONE such best-effort call per deliberate keypress, and only when actually toggling (the round-1 \`pane layout --current\` query was removed), so Z is lighter than the existing picker path. host\_zoom discards the error (best-effort): a failing or exited herdr just leaves the in-plugin zoom working. A herdr wedged long enough to block Command::output means the host multiplexer itself is wedged, which compromises the whole session regardless of this key. A bounded-timeout subprocess wrapper is a repo-wide concern (it applies equally to the picker and every herdr call) needing a new dependency or a worker thread, out of scope for this feature; logged as a possible follow-up for a global herdr-call timeout policy.

smarzban:
## Orchestrator review — Approve (Round 2)

**What this PR does:** adds `Z` (Shift+`z`), a full-screen toggle. The first press opens the selected file like `Enter` and zooms the viewer's herdr pane to fill the whole terminal; a second press (or `Esc`/`q`, or `z`) returns to the split. Read-only: the pane zoom is a host layout op, and it degrades to the in-pane zoom when there is no live herdr.

**Round 1 (discovery, 4 holistic seats + lens-security + lens-tests + deterministic scan) surfaced a real cluster**, all around one root cause: the first cut introduced host-zoom state that only `Z` managed, and it decided the toggle by querying `herdr pane layout` on every press. That produced: teardown leaks (quit / `Esc` / re-root left the host pane zoomed), a `z`/`Z` desync (the split redrew inside a still-full-screen pane), and a toggle that could not toggle back when herdr was absent (the query returned `None`, so a second `Z` re-activated instead of un-zooming).

**Fixed (cumulative):**
- Reworked `Z` to key off an owned `host_zoomed` flag instead of a per-press herdr query. It now toggles correctly with or without a live herdr.
- Every exit path releases the viewer's own host zoom through a single `leave_host_zoom` hook: `Esc`/`q`, `z`, a worktree re-root, and quit. The host pane no longer lingers full-screen behind a two-column split.
- `Z` now makes at most one best-effort herdr call per deliberate press (the `pane layout` query is gone), lighter than the existing worktree-picker path.
- Added teardown tests (Close / `z` / re-root each issue `--off`), a real `Z`->`Z` round-trip test, a without-herdr toggle test, and `LiveHerdr::run` seam tests. Corrected the `host_zoomed` doc to describe the adopt-on-`Z` behavior honestly.

**Dismissed (2, both held by an independent dismissal-audit on a different model):**
- *Blocking herdr subprocess on the input thread, no timeout.* Pre-existing accepted pattern (the worktree picker at `picker.rs:162-163` does the same), now one best-effort call per deliberate keypress. A global herdr-call timeout policy is a reasonable repo-wide follow-up, out of scope here.
- *`Z` can adopt-then-release a pane the user zoomed with herdr's own `prefix+z`.* A deliberate, documented trade-off of the flag-based toggle (which is what lets it work with no herdr); only manifests when mixing `prefix+z` with `Z` on the viewer pane, and the outcome is reasonable.

**Deferred / not covered:** a global bounded-timeout policy for herdr subprocess calls (follow-up); the `prefix+z`-then-`Z` adopt edge (documented, accepted). Live end-to-end behavior was confirmed manually on herdr 0.7.3.

**Decision: Approve** (agrees with the Round-2 PASS). Not merging here: `main` is protected and the merge is the maintainer's call.

PR #95, opened 2026-07-12
feat: configurable tree/content layout (tree_width, tree_position, tree_max_cols)

smarzban:
## What

Two read-only `config.toml` keys for the tree/content layout inside the viewer's own pane, layered on the existing settings-config surface (same loader, `config > default` precedence, defensive parse, display-only Settings overlay):

- **`tree_width`** — the directory tree column's share of the viewer pane, a percent clamped to `20..=80`. Seeds the **startup** split; the live grow/shrink keys and divider drag still adjust it within a session.
- **`tree_position`** — `"left"` (default, today's layout) or `"right"`. Lenient: trimmed, case-insensitive; an unrecognized value falls back to `"left"`.

**The default tree width changes 40 → 30** (owner-decided), giving the content pane a little more room out of the box. Default side stays left.

Not in scope: the herdr-level pane split (~50%, the host decides that; `plugin pane open` exposes no ratio), and persisting the live-adjusted split (config seeds startup only, consistent with the read-only design).

## How

- `Config.tree_width: Option<u32>` + `Config.tree_position: Option<String>` → `resolve()` → `EffectiveSettings.tree_width: u16` + `.tree_position: TreePosition` (clamp + lenient map, no env var).
- Split-range constants (`DEFAULT_TREE_WIDTH=30`, `MIN=20`, `MAX=80`) move to `config.rs` as one source of truth; the controller's `SPLIT_DEFAULT/MIN/MAX` point at them so the config clamp and the live-resize clamp can't drift.
- `Controller` gains a `tree_position` field (carried across a re-root, like `split_pct`) + `apply_tree_width` / `apply_tree_position` startup appliers (mirroring `apply_hide_dotfiles` / `apply_scroll_lines`), wired in `app::run`.
- The presenter's `columns()` — the single helper both `draw()` and `geometry()` derive from — swaps the content/tree rects for the right side, so hit-testing and scrollbars follow automatically. Only `resize_split_to_col` also needs to know the side (so a drag "grows the tree" on either side).
- Effective values shown read-only in the `?` overlay's Settings section.

Zero new Cargo dependencies.

## Tests

Test-first, one atomic commit per task (T-1..T-6). Resolver clamp/lenient cases, controller appliers + re-root carry, presenter geometry for both sides, per-side divider drag, settings display row, and a docs-consistency drift guard. Full suite green; `cargo fmt --check` / `clippy -D warnings` / `cargo build --release` / `cargo audit` clean.

## Docs

README config block + precedence list + roadmap line, a Layout section in `config.example.toml`, CHANGELOG `[Unreleased]` (`Added` the two keys, `Changed` the 40→30 default), and CONTEXT.md glossary — all in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ All notable changes to this project are documented here. The format is based on
 ## [Unreleased]
 
 ### Added
+- **Configurable tree/content layout via `tree_width`, `tree_position`, and `tree_max_cols` keys in
+  `config.toml`.** `tree_width` sets the directory tree column's share of the viewer pane (a percent
+  from `20` to `80`; the content pane takes the rest), `tree_position` puts the tree on the `"left"`
+  (default) or `"right"` of the content pane, and `tree_max_cols` caps the tree at a maximum number
+  of **columns** so it does not balloon on a very wide pane — the tree is drawn at
+  `min(tree_width% of the pane, tree_max_cols)`. These seed the **startup** split inside the viewer's
+  own pane (not the size of the herdr pane, which the host decides); you can still resize the split
+  live with the grow/shrink keys or by dragging the divider, and "grow" always widens the tree
+  whichever side it is on. `config > default`, no environment variable; an out-of-range width clamps
+  into `20..=80`, an unrecognized position falls back to `"left"`, and the column cap clamps into
+  `10..=1000` (set it high to disable). All three effective values are shown in the `?` overlay's
+  Settings section.
 - **Configurable mouse-wheel scroll speed via a `scroll_lines` key in `config.toml`.** Sets how many
   lines each wheel event advances the content pane (and how many items it moves in the file-search
   list, and lines in the `?` help overlay); the directory tree still moves one row per event.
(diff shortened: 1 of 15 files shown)


smarzban:
## Empanel gate — PASS (converged)

Multi-model review of the full combined change (`tree_width` + `tree_position` + `tree_max_cols`), run on the squashed branch.

**Panel:** deterministic scan + holistic ×4 (`glm-5.2`, `gpt-5.5`, `opus-4-8`, `deepseek-v4-pro`) + `lens-spec` ×2. Deterministic scan clean both rounds. An independent dead/duplicate/over-engineering pass came back clean on all three axes (removed a redundant test-only accessor + a stale comment).

### Round 1 — findings (1 medium, 4 low) — all fixed

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | **medium** | Manual resize jumped a capped/narrow tree to a fixed 10% of the pane on very wide (>~300 col) panes | Split floor is now **pane-aware** (`min_tree_split_pct` = the % yielding ≥ `MIN_TREE_MAX_COLS` columns), shared by `columns()` and the interactive resize, so a narrow/capped tree is representable on any pane width |
| 2 | low | `engage_manual_split` underestimated the tree width by ~2 cols when a tree scrollbar is drawn | Account for the 2-col scrollbar gutter when reconstructing the outer width |
| 3 | low | No test for `tree_position = right` combined with the cap biting | Added a Right-position + cap geometry test |
| 4 | low | AC-4 malformed test didn't prove the *whole* config defaults | Fixture carries a valid `hide_dotfiles` and asserts it is dropped |
| 5 | low | AC-17 same as AC-4 | Same fix |

### Round 2 — verification (single model)

Scoped to the prior findings + the fix diff: **0 findings**, no regression. Deterministic scan re-run clean.

**Verdict: PASS.** No gating findings remain. CI 7/7 green; the 14 build/fix commits were squashed to one (tree-identical).

PR #102, opened 2026-07-15
fix: normalize Shift+lowercase chars for Windows terminals

### LINKED ISSUE #85: Shift+O and Shift+R not recognized on Windows — key events passed through to terminal
**Is this a reproducible bug?**
- I confirm this is a reproducible bug

## Current behavior

Pressing `Shift+o` (Open with default app) and `Shift+r` (Reveal in file manager) types the letters into the terminal instead of triggering the file viewer actions. The `e` key (open in editor) works fine because it's lowercase and doesn't involve Shift.

## Expected behavior

- `Shift+o` should open the selected file with the OS default application.
- `Shift+r` should reveal the selected file in Windows Explorer.

## Root cause

In `src/input.rs`, the `map_key()` function:

1. Correctly filters out modifier chords (Ctrl/Alt) but allows Shift:
```rust
if key.modifiers.difference(KeyModifiers::SHIFT) != KeyModifiers::NONE {
    return None;
}
```

2. Only maps uppercase characters:
```rust
KeyCode::Char('O') => Some(Intent::OpenWithApp),
KeyCode::Char('R') => Some(Intent::RevealInFileManager),
```

3. On Windows terminals, `Shift+o` is reported as `KeyCode::Char('o')` with `KeyModifiers::SHIFT` — not `KeyCode::Char('O')` with no modifiers. The lowercase `Char('o')` with `SHIFT` falls through to `_ => None` and the keystroke passes through to the terminal.

## Steps to reproduce

1. Open file viewer: `prefix+f`
2. Select a file in the tree
3. Press `Shift+o` — letter "O" is typed in the terminal
4. Press `Shift+r` — letter "R" is typed in the terminal
5. `e` (lowercase) works correctly

## Suggested fix

Handle both uppercase and `SHIFT`+lowercase in the match:

```rust
KeyCode::Char('O') | KeyCode::Char('o')
    if key.modifiers.contains(KeyModifiers::SHIFT) =>
    Some(Intent::OpenWithApp),
KeyCode::Char('R') | KeyCode::Char('r')
    if key.modifiers.contains(KeyModifiers::SHIFT) =>
    Some(Intent::RevealInFileManager),
```

Or normalize shifted characters to uppercase before matching.

## Environment

- OS: Windows 11
- herdr version: 0.7.2
- herdr-file-viewer version: 1.9.0
- Terminal: Windows Terminal
- Shell: PowerShell 5.1

sanirudh17:
## Problem

On Windows terminals, pressing Shift+O sends KeyCode::Char('o') with SHIFT modifier instead of KeyCode::Char('O') with no modifier. The same applies to Shift+R. This caused the OpenWithApp (O) and RevealInFileManager (R) keybindings to fail silently on Windows.

## Root Cause

In src/input.rs, the decode() function stripped the SHIFT modifier before lookup but did not normalize the character case. Since the registry binds uppercase Char('O') and Char('R'), the lowercase variants from Windows terminals never matched.

## Fix

Added normalization in decode(): when SHIFT is the only modifier and the character is lowercase ASCII, convert to uppercase before lookup. This is a pure function change with no side effects.

## Test Results

All tests pass (308 unit tests, 259 integration tests):

- **New test**: windows_shift_lowercase_normalizes_to_uppercase_for_lookup — validates Char('o') + SHIFT → OpenWithApp and Char('r') + SHIFT → RevealInFileManager
- **Existing tests**: All 38 input tests pass, confirming no regression
- **Full suite**: 308 unit + 259 integration tests pass (1 pre-existing unrelated failure in markdown rendering)

Closes #85

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -76,11 +76,25 @@ pub(crate) fn default_bindings() -> EffectiveBindings {
 /// typing — not a chord — and some terminals report them with the Shift bit set. `key.code` is the
 /// already-normalized lookup key: a shifted char carries its case in `key.code`, and a named key
 /// reports its base [`KeyCode`] with Shift in the modifiers (stripped by the guard below).
+///
+/// On Windows, some terminals report `Shift+o` as `Char('o')` with `SHIFT` modifier rather than
+/// `Char('O')` with no modifier. When SHIFT is the only modifier and the character is lowercase,
+/// normalize to uppercase before lookup so shifted keys resolve correctly.
 pub(crate) fn decode(key: KeyEvent, bindings: &EffectiveBindings) -> Option<Intent> {
     if key.modifiers.difference(KeyModifiers::SHIFT) != KeyModifiers::NONE {
         return None;
     }
-    bindings.intent_for(key.code)
+    // Normalize SHIFT+lowercase to uppercase for cross-platform compatibility.
+    // On Windows terminals, Shift+o may arrive as Char('o') + SHIFT instead of Char('O').
+    let code = if key.modifiers.contains(KeyModifiers::SHIFT) {
+        match key.code {
+            KeyCode::Char(c) if c.is_ascii_lowercase() => KeyCode::Char(c.to_ascii_uppercase()),
+            other => other,
+        }
+    } else {
+        key.code
+    };
+    bindings.intent_for(code)
 }
 
 /// Borrow the process-lifetime default [`EffectiveBindings`], built once from the [`REGISTRY`].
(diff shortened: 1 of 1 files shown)


smarzban:
Validated this on Windows using Windows Terminal and PowerShell.

   - herdr-file-viewer v1.9.0: Shift+O and Shift+R both failed.
   - herdr-file-viewer v1.13.0: Shift+O opened the default app and Shift+R opened Explorer, with no input leakage.

This appears fixed in a later release, so I’m closing this as resolved.

PR #101, opened 2026-07-15
feat: add file annotations

mschwarzmueller:
## Summary

Adds session-only, read-only annotations for files and source line/ranges. The feature is intended for reviewing code and collecting structured instructions that can be copied into a prompt for an AI agent.

Annotations never modify files or git state, are held only in memory, and are scoped to the viewer's current root.

Example clipboard output:

```text
<file-annotations>
- README.md:Clarify the fallback.
- src/app.rs:42:Explain the ignored result.
- src/controller/mod.rs:42-47:Why is this guarded twice?
</file-annotations>
```

### Annotation workflow

- Press `a` on a selected file to add a file-level annotation.
- In content-focused line-select mode (`L`), press the mode-local `a` to annotate the selected line/range.
- Press `A` to open the annotation overview:
  - `j`/`k` or `↑`/`↓`: move
  - `Enter`/`e`: edit
  - `d`: delete one
  - `D`: clear all
  - `y`: copy the deterministic `<file-annotations>` export through OSC 52
  - `Esc`/`q`: close
- The global `add_annotation` and `show_annotations` actions can be remapped through the existing `[keys]` configuration. Modal-local controls remain fixed.

### Visual feedback

- A nonzero annotation count appears on the content border.
- Annotated files show `@` in the tree without changing row width or hiding the git marker.
- The applied content title is prefixed with `@`.
- Source-mapped line/range annotations receive a subtle background, including a bounded cue for blank lines.
- Active line-select, mouse-selection, and search styles take precedence over the persistent annotation background.
- Overview and Add/Edit dialogs use stable, bounded widths; long notes, rows, and validation errors do not resize them horizontally.

### Architecture and safety

- Adds a typed `AnnotationStore` with immutable root-relative targets, normalized 1-based ranges, stable identities, deterministic ordering, and canonical serialization.
- Annotation state remains Controller-owned and is projected into an owned `ViewState`; `TreeModel` and `Node` remain annotation-free.
- Numeric line indicators are projected only for the applied file when `RenderResult::source` is present.
- Rendered Markdown and diff views keep file/title markers but intentionally do not guess source-line mappings.
- A successful root/worktree change clears annotations; failed and same-root changes retain them.
- Reuses the existing Unicode-safe single-line prompt editor and clipboard seam.
- No new dependency, persistence layer, configuration setting, renderer behavior, or file/git mutation.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --all-targets -- -D warnings`
[...]

PR #99, opened 2026-07-14
feat: configurable content-preview caps (preview_max_lines / preview_max_kib)

smarzban:
## What

Makes the content-pane truncation caps configurable instead of the fixed 5000-line / 1 MB. Two new `config.toml` keys (`config > default`, no env tier), each clamped into a safe range:

| Key | Default | Range | Caps |
|---|---|---|---|
| `preview_max_lines` | 10000 | 100–100000 | lines before a truncated preview |
| `preview_max_kib` | 1024 (= 1 MB) | 64–65536 (64 MB) | size before truncating; also bounds the disk read |

Truncation still fires on whichever cap is hit first, and now applies uniformly to file content **and** large diffs. The `⚠ Truncated preview…` notice reports the actual configured cap instead of a hardcoded "1 MB / 5000-line".

## Why

Answers a direct ask: files past 5000 lines were cut with no way to raise the limit. The default line cap is also bumped 5000 → 10000 (friendlier for large-but-normal-width source, still reachable under the unchanged 1 MB byte cap).

## Design notes

- **Safety preserved (AC-N1).** The byte cap still bounds the actual file read; the upper clamp (64 MB) keeps every setting a *finite* read, so no configured value can make the viewer slurp an arbitrary file whole. Default stays 1 MB.
- **Single wiring seam.** `render::Caps` is injected into `classify` / `render`, resolved from config via `EffectiveSettings::preview_caps()` (the one place KiB → bytes). A lockstep test pins `Caps::default()` to a config-absent `resolve()` so the two default sources can't drift.
- Effective values surface in the `?` overlay's Settings tab.

## Docs (same PR)

`docs/configuration.md`, `config.example.toml`, `CHANGELOG.md` (`[Unreleased]`), plus the Settings-overlay rows and drift-guard tests (`tests/docs_consistency.rs`).

## Tests

- New: injected-cap truncation (a file the default shows whole is truncated at a smaller configured cap — line and byte paths), config parse → resolve → clamp for both keys, TOML key-name binding, and the default-lockstep guard.
- Existing diff-truncation tests made cap-relative so they can't rot on a future default change.
- Deterministic tier green locally: `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, full `cargo test` (0 failures), `cargo audit` unchanged.

Read-only, no new deps.

PR #103, opened 2026-07-15
feat: show effective values in the Settings help tab

smarzban:
## Summary

The `?` overlay's Settings tab printed `(default)` for every command left unconfigured, so it
couldn't answer the question it exists for: what is actually in effect. Every row now shows its
resolved value, after config/env/default precedence.

The renderer commands (`markdown`, `diff`, `syntax`) are not listed. They are set in the config file
and documented there, and their built-in values are long argv (markdown's `-s` resolves to an
absolute path into the install dir) that wrapped across rows, broke the `=` column, and leaked the
build location into the overlay. The editor and opener rows stay: short, and the command is the
point of those rows.

## Changes

- `help::settings_text` shows each effective value; the `(default)` placeholder is gone.
- The `markdown`/`diff`/`syntax` rows are dropped; `help::SettingsWired` carries only the editor and the opener labels.
- `opener::default_opener_display` gives the `open`/`reveal` rows a per-OS label when unset.
- Every row is one line, so the `=` column holds at any pane width.
- Docs in the same PR: CHANGELOG entry, `docs/configuration.md`, and the CONTEXT glossary.
- Also here, both unrelated to the feature and separately committed:
  - the About version fixture bump to v1.12.0 (cosmetic carry-forward, missed on #99 and #101; the live line uses `env!(CARGO_PKG_VERSION)` so it never failed CI)
  - `docs/specs/` added to `.gitignore`, beside the existing `/specs/` and `/docs/adr/` entries

## Test plan

- [x] `cargo test` is green (1088 passed)
- [x] Docs updated in this PR (CHANGELOG + `docs/configuration.md`) for any user-facing change
- [x] Live-tested in a herdr split pane: 11 rows, each on one line, no wrap, no absolute path


smarzban:
## Empanel — Round 1
🚫 **BLOCK** — 4 blocking finding(s) must be resolved or justified.

Findings: 5 total — 0 critical · 1 high · 3 medium · 1 low · 0 info.
_Reviewed by:_ holistic + lens-security + lens-tests · models: anthropic:claude-opus-4-8, deepseek:deepseek-v4-pro, ollama:glm-5.2, openai-codex:gpt-5.5
⚠️ **Coverage:** 12/13 planned reviewer passes voted — lost: lens-spec (spec dropped from the PR and describes superseded (d)-marker behavior)

### Must fix
- 🔴 **[HIGH]** \[test-coverage\] default\_opener\_display has zero tests across all 6 OS/action branches — `src/opener.rs:75` · 4/4 models · _by:_ holistic, lens-tests _(test-coverage)_
  This is new public API with 6 branches (Mac Open, Mac Reveal, Linux Open+Reveal, Windows Open, Windows Reveal). The Windows branches read %SystemRoot% via system\_root() and format Explorer paths. No unit or integration test exercises any branch. If a future refactor changes any branch (e.g. Mac Reveal drops \`-R\`, Windows Reveal formats \`/select,\<path\>\` differently), it would silently show wrong opener info in the Settings tab.
  _Fix:_ Add unit tests in the opener.rs test module covering all 6 (os, action) pairs, mirroring the existing \`opener\_argv\` tests. For the Windows paths, test with a known \`%SystemRoot%\` value set via \`std::env::set\_var\` in a test (or refactor to accept an explicit \`system\_root\` parameter for testability).
- 🟠 **[MEDIUM]** \[test-coverage\] Live SettingsWired startup wiring is untested — `src/app.rs:100` · 4/4 models · _by:_ lens-tests _(test-coverage)_
  The tests exercise \`settings\_text\` with hand-built \`SettingsWired\`, but not the \`app::run\` wiring that computes it from the same editor/opener resolution used by the live components. Swapping open/reveal, omitting the platform editor fallback, or passing stale wired values to the controller would not be caught by the current pure-helper tests.
  _Fix:_ Extract a pure \`SettingsWired\` builder from \`(EffectiveSettings, platform\_editor, OsKind)\` and test config/env/default precedence and open/reveal pairing, or add a startup wiring test that opens the Settings tab after applying a config and asserts the displayed rows.
- 🟠 **[MEDIUM]** \[test-coverage\] Settings command precedence branches are untested — `src/help.rs:245` · 3/4 models · _by:_ lens-tests _(test-coverage)_
  The new Settings output is tested for a configured editor plus wired open/reveal defaults, but not for configured open/reveal overrides or the \`eff.editor == None\` with \`wired.editor == Some\` platform-default path. A regression that always displayed opener defaults or showed \`unset\` instead of a resolved platform editor would still pass.
  _Fix:_ Add \`settings\_text\` tests with \`eff.open\`/\`eff.reveal\` set and assert their exact rows override \`wired\`, plus a test with \`eff.editor: None\` and \`wired.editor: Some(...)\` asserting the editor row shows the wired value.
- 🟠 **[MEDIUM]** \[test-coverage\] settings\_text open/reveal rows with config overrides are untested — `src/help.rs:262` · 2/4 models · _by:_ holistic, lens-tests _(test-coverage)_
  The \`eff.open.as\_ref().map(\|argv\| argv.join(" "))\` and corresponding \`eff.reveal\` paths are never exercised. All tests use \`sample\_eff()\` (open/reveal=None) or \`Config::default()\` resolution (also None). If the \`join(" ")\` rendering for user-configured opener commands were broken (e.g. missing space separator, or empty-vec edge case), no test would catch it.
  _Fix:_ Add a test with \`eff.open = Some(vec!\["myopen".into(), "-x".into()\])\` and assert the exact row \`open = myopen -x\`. Also test the empty-vec edge case if it should be handled.

### Advisory (non-blocking)
- ⚪ **[LOW]** \[test-coverage\] settings\_text\_lists\_every\_setting\_row does not assert exact confirm\_discard = off row value — `src/help.rs:798` · 1/4 models · _by:_ lens-tests _(test-coverage)_
  The test checks \`text.contains("confirm\_discard")\` in the key-iteration loop but the exact-row assertion list omits \`"confirm\_discard = off"\`. With \`sample\_eff().confirm\_discard = false\`, the rendered value should be \`off\`, but only a substring check guards it. The \`on\` value is properly tested as an exact row in \`settings\_text\_shows\_builtin\_default\_scalars\` but the \`off\` case relies on a weaker assertion.
  _Fix:_ Add \`"confirm\_discard = off"\` to the exact-row assertion list in \`settings\_text\_lists\_every\_setting\_row\`.

smarzban:

**Panel:** holistic ×4 (glm-5.2, gpt-5.5, opus-4-8, deepseek-v4-pro) + `lens-tests` ×4 + `lens-security` ×4 + the deterministic scan. Two seats non-voted on the first attempt (glm-5.2 holistic hit its turn cap; gpt-5.5 lens-security hit a provider reset) — both were re-run and voted, so the panel is whole. `lens-spec` was skipped and is recorded in Coverage: the spec was dropped from this PR and describes the superseded `(d)`-marker behaviour, so conformance is **not** checked this round.

**Lens decision:** fired `lens-tests` (deterministic `suggest` flagged "source changed with no added test lines") and `lens-security` (the diff touches `opener.rs`, whose neighbours shell out — a prior gate caught a HIGH Windows `explorer` hijack in this exact file). Skipped privacy/contracts/migrations/subtle-correctness/simplify/performance/infra: no trigger matched. Security came back 4/4 clean.

**What I agree with and have fixed** (`8a274a6`):

1. **`default_opener_display` was untested** (high, 4/4 models). It hand-mirrors `opener_argv`'s program and fixed flags, while `opener_argv` itself has six per-branch tests. Nothing tied the two together, so the overlay could drift into advertising a command that isn't what `O`/`R` spawn. Every branch is now pinned to `opener_argv`'s argv[0], including the Windows `%SystemRoot%` resolution and the `/select,` flag, plus a guard that Open and Reveal can't be crossed.
2. **An empty override rendered blank** (medium, 2/4, contested — and real). I confirmed it end to end rather than from the title: `open = ""` resolves to `Some([])` (probed directly), and the existing `with_overrides_empty_prefix_falls_through_to_the_per_os_default` test proves `CommandOpener` then falls through to the per-OS default. The row joined that empty argv and showed nothing, so the display contradicted the behaviour in exactly the case the code deliberately handles. It now reports the same fallthrough.
3. **The `SettingsWired` startup wiring was untested** (medium, 4/4). It lived inside `run`, which needs a terminal. Extracted to `settings_wired`, now covered — the labels can't be crossed and the editor follows config > `$EDITOR` > platform default.
4. **Untested `settings_text` branches** (medium, 3/4, contested). The wired-editor fallback and the configured-opener rows now have tests.

**Dismissed:** nothing. All four gating clusters were real on inspection.

**Advisory, not blocking:** the lone low finding that `settings_text_lists_every_setting_row` doesn't assert an exact `confirm_discard = off` value (1/4 models). The row's presence and value are already covered by the exact-row assertions.

`cargo test` 1095 green (7 new), fmt/clippy/audit clean.

**Decision: 🔴 Request changes** — fixes are pushed; Round 2 verifies them.

smarzban:
## Empanel — Round 2

_Reviewed by:_ verify · models: anthropic:claude-opus-4-8
⚠️ **Coverage:** 1/2 planned reviewer passes voted — lost: lens-spec (spec dropped from the PR and describes superseded (d)-marker behavior)

### Progress since Round 1
✅ Resolved (4): \[test-coverage\] default\_opener\_display has zero tests across all 6 OS/action branches; \[test-coverage\] Live SettingsWired startup wiring is untested; \[test-coverage\] Settings command precedence branches are untested; \[test-coverage\] settings\_text open/reveal rows with config overrides are untested — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

smarzban:

**Verdict: pass.** Round 2 was single-model verification (opus-4-8) of Round 1's four findings against the fix commit, plus the full deterministic scan. All four verified resolved, nothing still blocking, no regression. The verifier returned a real vote (no near-cap, no warning), so the clean result is coverage, not silence.

### What this PR implements

The `?` overlay's Settings tab used to print `(default)` for every unconfigured command, so it couldn't answer what's actually in effect. It now lists each setting's resolved value, one row per line. The renderer commands (`markdown`/`diff`/`syntax`) are not listed at all: they're set in the config file and documented there, and their built-in argv is long enough (markdown's `-s` resolves to an absolute path into the install dir) that showing it wrapped rows, broke the `=` column, and leaked the build location.

### Fixed (cumulative)

- `default_opener_display` had zero tests while the `opener_argv` it mirrors has six; every branch is now pinned to `opener_argv`'s argv[0], including the Windows `%SystemRoot%` resolution and the `/select,` flag.
- An empty override (`open = ""`) rendered a blank row while `CommandOpener` actually falls through to the per-OS default. The row now reports that fallthrough.
- The `SettingsWired` startup wiring was untestable inside `run`; extracted to `settings_wired` and covered, so the open/reveal labels can't be crossed and the editor still follows config > `$EDITOR` > platform default.
- The wired-editor fallback and configured-opener rows gained tests.
- Also carried: the About version fixture bump to v1.12.0, and `docs/specs/` added to `.gitignore`.

### Deferred / not covered

- **`lens-spec` did not run** (shown in Coverage). The Agent SDLC spec was dropped from this PR and describes the superseded `(d)`-marker behaviour, so spec conformance is unverified. The behaviour was instead checked live in a herdr pane.
- **Advisory, not fixed:** one low finding (1/4 models) that `settings_text_lists_every_setting_row` doesn't assert an exact `confirm_discard = off` value. The row and its value are already covered by the exact-row assertions.
- **`app::run` itself** remains uncovered end to end (it needs a terminal). The wiring it performs is now tested; the loop around it is not.
- The renderer *arguments* are no longer visible anywhere in-app. Deliberate: `docs/configuration.md` is the reference.

`cargo test` 1095 green · fmt/clippy/audit clean · full CI matrix green · live-tested in a herdr split pane.

**Decision: ✅ Approve**

PR #104, opened 2026-07-15
feat: render binary documents (docx/odt/pdf/pptx/xlsx)

Javamomma:
## What

Office and PDF files (`.docx`, `.odt`, `.pdf`, `.pptx`, `.xlsx`) previously showed the `[binary file: preview not shown]` placeholder. They now **render** in the content pane, converted to text by a best-in-class external tool and shown through the existing markdown/`glow` pipeline — so headings, lists, and tables render and wrap to the pane like a `.md` file.

| Kind | Converter | Output |
| --- | --- | --- |
| `.docx`, `.odt` | `pandoc … -t gfm` | markdown (stdout) |
| `.pdf` | `pdftotext -layout` | text (stdout) |
| `.pptx`, `.xlsx` | `libreoffice --headless --convert-to …` | text / CSV (temp file) |

## How it fits the existing design

- **Delegate, don't reinvent (ADR-0001).** Converters are optional runtime deps (like `glow`/`delta`/`bat`), injected via `Renderers` so tests stay hermetic. A missing or failing converter degrades to a short notice naming the tool to install — never a crash or empty pane (AC-24/25 parity).
- **New pieces:** a pure `document` module (extension → `DocKind`, and a `Converter` spec that is either *stdout* or *temp-file* since LibreOffice only writes files); `render::render_document` (run the converter on the path, then render its output through the markdown delegate); a `RenderedDocument` view mode. A document is binary, so it defaults to — and only offers — the converted view (a diff/syntax view of its bytes would be meaningless).
- **Wrapping:** the pane width threads to `glow` for documents exactly as for markdown, and a document re-flows on pane resize / `w` toggle. The transformed-view left inset applies too.

## Security

These tools read a *binary* file, so — unlike the stdin-fed renderers — the converter receives the file **path** (the one file type that can't be piped on stdin). The path is already confined to the tree root by `classify`'s caller, and the converter's output is re-neutralized by `to_text` before display, so the escape-neutralization trust boundary (AC-27) is unchanged. Temp-file converters run in a fresh per-invocation temp dir that is removed afterward.

## Scope / follow-ups

Converters are built-in defaults (not user-configurable) in this PR, per YAGNI — happy to add `config.toml` overrides in a follow-up if you'd like them, mirroring the `markdown`/`diff`/`syntax` keys. `install-renderers.sh` isn't extended here; `docs/renderers.md` lists the manual installs.

## Tests / checks

- New unit tests (`document` kind detection + converter specs; `view_policy` document precedence).
- New **hermetic** e2e tests for `render_document` (stub converters + `cat` as the delegate): stdout-converter output reaches the pane, a missing converter yields a notice not a crash, and a temp-file converter's written output is read back and rendered.
- `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, and the full `cargo test` suite are green locally.
- Docs updated in this PR per the contributor guide: `CHANGELOG.md`, `docs/renderers.md`, `docs/usage.md`, `ARCHITECTURE.md`.


smarzban:
Thanks for the thoughtful implementation and tests. I’m going to close this because automatic document conversion adds more performance and cross-platform complexity than I want for the viewer right now. I appreciate the contribution.

PR #110, opened 2026-07-17
feat: git-status mode (d)

jellydn:
## What

Add a sticky **`d` git-status mode**:

- Filter the tree to **current working-tree status** only (`M` / `A` / `?` / `D`) — independent of baseline
- Force **working-tree diffs** in the content pane (file or directory-scoped, via `delta`)
- Sticky until `d` again
- **Mutually exclusive** with baseline-aware `c` (branch/history review)
- `b` still updates the stored baseline while `d` is on; content stays working-tree

## Why

`c` is baseline-aware (merge-base vs `HEAD`) and is useful for reviewing a whole branch. Reviewers also need a mode that always means “what’s dirty right now,” with diffs vs the working tree — including a whole-directory working-tree diff when a folder is selected.

Directory-level “scan everything under this folder” belongs inside that status mode, not as a separate product feature on `d`.

## How

- New `Intent::ToggleStatusMode` bound to `d` (registry + docs)
- Cached `git_status` separate from the baseline-dependent `changed` set
- Tree filter reuses the existing changed-only path, fed from `git_status` while status mode is on
- Render path forces `ViewMode::Diff` + `Baseline::Head`
  - file → `GitService::diff`
  - directory → `GitService::diff_directory` (pathspec-scoped; empty pathspec at root)
- Async worker path (AC-23) unchanged for files; directory jobs set `directory_diff`
- Entering `d` turns off `c` and vice versa; `b` recomputes the baseline changed-set without leaving `d`

### Commits

1. `feat(status-mode): add sticky d mode for working-tree status`
2. `test(status-mode): cover filter source, exclusivity, and WT diffs`
3. `docs(status-mode): document d in usage, keys, config, and changelog`

### Checklist

- [x] Behaviour tests (filter source, exclusivity, sticky `b`, file/dir Head diffs)
- [x] Registry + docs drift guards
- [x] Docs: usage / keys / configuration / CHANGELOG Unreleased
- [x] No breaking change to `c` / `b` outside status mode

### Testing

```bash
cargo test --lib
cargo test --test controller status_mode
cargo test --test controller_async status_mode
cargo test --test docs_consistency
cargo fmt --check
cargo clippy --all-targets -- -D warnings
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project are documented here. The format is based on
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html). Entries are short on purpose; follow the
 `→` links for the full detail.
 
+## [Unreleased]
+
+### Added
+- Git-status mode (`d`): filter the tree to current working-tree status and force working-tree diffs (file or directory-scoped); sticky until `d` again, mutually exclusive with baseline-aware `c`. → [usage](docs/usage.md#git-awareness) · [keys](docs/keys.md)
+
 ## [1.13.0] - 2026-07-16
 
 ### Added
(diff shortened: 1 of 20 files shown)


smarzban:
Thanks @jellydn for the thoughtful PR, this is a really useful addition for reviewing current working-tree changes.
I added a small follow-up commit so directory-level diffs also include untracked files, and fixed a race in the async directory-diff test. The existing status-mode behavior is unchanged. 🙌🏻

PR #111, opened 2026-07-18
Add D key: cycle Diff/FullDiff between delta unified, delta side-by-side, and raw git diff

rrrrnmtsu:
## Summary
- `D` (Shift+`d`) now cycles the Diff/FullDiff views through three renders: `delta` unified (current default) -> `delta --side-by-side` -> plain, unstyled `git diff` text -> back to unified.
- `delta --side-by-side` is spawned with stdout piped (not a real tty), so it can't auto-detect a terminal width and its two columns were sized to a stale fallback regardless of the pane. A second width channel (`pane_width`, unconditional — unlike the existing markdown-only `wrap_width`, which is gated by the wrap preference) now feeds delta's own `--width` for Diff/FullDiff, mirroring the existing markdown fit-to-pane path, and reflows live on resize (new `rerender_after_resize`, refactored out of the old markdown-only `rerender_markdown_reflow`).
- New `Intent::ToggleDeltaRaw` + `DiffRenderMode` enum (`Delta` / `DeltaSideBySide` / `Raw`), threaded through `RenderJob` and `ContentProvider::render_at_width`. `Raw` uses `cat` as the delegate rather than skipping rendering outright, so the existing renderer/timeout/fallback machinery stays the single code path for every mode.
- Docs (`docs/keys.md`, `docs/configuration.md`) and the keybinding registry/tests updated to match — `cargo test`'s doc-sync tests (`configuration_doc_lists_every_remappable_intent`, `keys_doc_table_documents_every_registry_action_ac21`, `every_intent_has_at_least_one_key`) pass.

## Motivation
There's currently no way to see a raw, unstyled diff (e.g. to paste one verbatim somewhere) or a side-by-side layout without editing `config.toml` and relaunching. Both are common enough asks to warrant a keyboard toggle.

## Test plan
- [x] `cargo build --release` — clean build
- [x] `cargo test --release` — all pass except `search_routes_keys_to_query_and_n_cycles_and_esc_restores`, which I verified also fails on unmodified `main` (a pre-existing flaky macOS pty-timing test, unrelated to this change)
- [x] Manually verified in a live herdr session: `D` cycles all three states with the expected notice text; side-by-side columns resize live when dragging the tree/content split or resizing the terminal

https://claude.ai/code/session_012coLDNNFQAf9H5GCLeLjmu


smarzban:
Thanks @rrrrnmtsu for the PR, this is a great feature and a really useful addition for switching between diff views. The implementation looks great.
I’ve added a few follow-up on top of your changes, including raw diff rendering, better resize/reflow handling, additional safety guards, and expanded tests and documentation.
Really appreciate the work on this!

PR #112, opened 2026-07-18
fix: support bash/zsh panes on Windows and open splits on the left

Grenghis-Khan:
## Summary
- Windows launchers detect `terminal.default_shell` and emit PowerShell (`& "…"`) or Git Bash/zsh (`"…"`) `pane run` forms, with a short `%USERPROFILE%\bin\hfv.exe` shim plus mirrored `assets/markdown-style.json` so glow styling still works.
- Split-pane openers (Windows + unix) place the Files pane on the **left** via `split --direction right` then `pane swap --direction left` (herdr has no left split).
- Docs/changelog/tests updated (Git 2.41+ note, winget renderer install, dual-shell + left-split guards).

## Test plan
- [x] Windows + PowerShell default shell: `prefix+f` opens Files on the left; glow markdown headings styled
- [x] Windows + Git Bash/zsh `default_shell`: same, no `&` parse error; launch is fast (shim)
- [x] macOS/Linux: `open-file-viewer` opens Files on the left of the work pane
- [x] Toggle still works (open → focus → close)
- [x] `cargo test --test launcher_content` (and Windows parse job if available)


smarzban:
Thanks @Grenghis-Khan for putting this together and testing it on Windows. The non-PowerShell support is useful, but the persistent shim, placement change, and conflicts with current launcher work make this too broad to merge safely. I’m going to close it, but I’ll consider a smaller PR focused only on shell compatibility, without the shim or placement change.

PR #105, opened 2026-07-16
fix: preserve UTF-8 in Windows launchers

mkioutcc:
## Summary

Fixes the native-Windows launchers opening the plugin install directory instead of the invoking pane's directory when `herdr pane list` contains non-ASCII titles or paths.

Windows PowerShell 5.1 decodes native-command stdout with the active legacy console code page. Herdr emits UTF-8 JSON, so non-ASCII text could be corrupted before `ConvertFrom-Json`; `Get-UserCwd` then caught the parse error and silently fell back to `$PluginRoot`.

## Changes

- Configure BOM-less UTF-8 for native-command output in both Windows launcher scripts before parsing `herdr pane list`.
- Apply the same UTF-8 setup to both inline Windows manifest actions before parsing `herdr plugin list --json`, covering non-ASCII install paths too.
- Add a blocking cross-platform regression guard that checks all four `ConvertFrom-Json` consumers initialize UTF-8 first.
- Document non-ASCII path/title support in `docs/windows.md` and add a `CHANGELOG.md` fix entry.

No new dependencies; read-only behavior is unchanged.

## Test plan

- [ ] `cargo test` is green
  - The new regression tests and PowerShell parse tests pass.
  - The full Windows run reaches the pre-existing `whats_new_body_is_rendered_via_markdown_renderer_when_present` failure; the same failure reproduces on an untouched `origin/main` worktree. Running the rest then reaches a local Windows elevation-policy error (`os error 740`) when starting the unrelated `update_banner` test binary.
- [x] `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo audit` pass
  - `cargo audit` reports only the repository's existing allowed `RUSTSEC-2026-0190` warning.
- [x] Docs updated in this PR (`CHANGELOG.md` + `docs/windows.md`)
- [x] `cargo build --release`
- [x] `cargo test --test launcher_content`
- [x] `cargo test --test launcher_ps1`
- [x] Live Windows smoke test for both split and tab launchers: each opened the invoking `Documents` directory correctly while the Herdr session contained non-ASCII pane titles and paths.

## Related

No issue filed; reproduced with Windows PowerShell 5.1 and herdr `0.7.4-preview`.

PR #114, opened 2026-07-20
fix: thread pane width into delta's -w for diff/full_diff

trsxxii:
## Summary

`delta`'s stdout is a plain OS pipe when this app spawns it (`Stdio::piped()`, not a tty), so
delta's own terminal-width auto-detection always fails and it falls back to a fixed, tty-less
width instead (observed as 80 columns for bare delta). `glow` already gets an explicit
`-w <pane_width>` for the same reason; `delta` (compact diff and full-file diff) never did,
based on the assumption that it "manages its own width" the way it would when attached to a
real terminal. This PR threads the measured pane width into delta's `-w` on every diff render,
and keeps it in sync across resizes.

**Verified against the real delta 0.19.2 binary** (not assumed from its docs) through the app's
actual render pipeline, at two pane sizes and both of delta's layout modes, before vs. after this
change: https://claude.ai/code/artifact/ed35d7fe-734c-4a71-a642-d8f7a631aafc

The effect is conditional on delta's own mode, which this app doesn't control (delta reads
`side-by-side` from the user's `~/.gitconfig` `[delta]` section directly — this app never passes
`--no-gitconfig`):
- **`side-by-side` on:** `-w` genuinely governs each column's wrap width. Confirmed by measuring
  the rendered column width directly: before, it's stuck at 40 (half of the 80-column fallback)
  regardless of pane size; after, it tracks the pane (~53 at a 140-wide pane, 30 at 90-wide).
- **`side-by-side` off (delta's default):** `-w` only changes the width of the decorative
  underline/overline rule (per `delta --help`); the diff content itself is never wrapped by it
  either way, on either side of this change — confirmed with a 175-character fixture line that
  stays on one row and is simply cut off at the pane edge in every capture.

## Changes

- `render_at_width` (`src/app.rs`) now injects `-w <width>` for `Diff`/`FullDiff`, reusing the
  same `with_wrap_width` helper already used for markdown/glow. `bat` is left alone (no `-w` flag,
  genuinely width-independent).
- Splits `diff_render_width()` (unconditional on the wrap toggle) from `md_wrap_width()` (gated on
  `effective_wrap()`), since delta's `-w` has nothing to do with the Presenter's wrap/h-scroll axis
  — a diff's wrap default is `false`, so reusing the gated helper meant the width never actually
  reached delta in practice.
- Adds `rerender_diff_reflow()` so a pane resize re-renders an open diff too, mirroring the
  existing markdown reflow path (and reusing the same worker backlog-collapse, so a rapid resize
  doesn't spawn a delta process per event).
- `toggle_wrap` (the `w` key) intentionally still only re-renders markdown: a diff's `-w` is
  independent of that toggle, so resending would be pure waste.
- Documents the `side-by-side` dependency in `docs/configuration.md`'s Known limitation note, and
  adds a `CHANGELOG.md` entry.

## Test plan

- [x] `cargo test` is green, including new tests: `render_at_width_points_diff_wrap_at_the_pane_width`,
      `render_at_width_points_full_diff_wrap_at_the_pane_width`,
      `render_at_width_leaves_diff_unmodified_without_a_pane_width`,
      `a_width_change_reflows_a_diff_at_the_new_width`, `toggle_wrap_on_a_diff_dispatches_no_render`
      (`cargo audit` reports only the repository's existing allowed `RUSTSEC-2026-0190` warning,
      same as PR #105)
- [x] Docs updated in this PR (`CHANGELOG.md` + `docs/configuration.md`)
- [x] Manual verification against the real `delta` binary: built both `main` (before) and this
      branch (after), same repo/fixture, at 140×40 and 90×35, in both delta layout modes — capture
      matrix and measured rule/column widths linked above

## Related

No issue filed; found while testing the diff pane at a wide split.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ All notable changes to this project are documented here. The format is based on
 ### Added
 - Git-status mode (`d`): filter the tree to current working-tree status and force working-tree diffs (file or directory-scoped); sticky until `d` again, mutually exclusive with baseline-aware `c`. → [usage](docs/usage.md#git-awareness) · [keys](docs/keys.md)
 
+### Fixed
+- Diff and full-file-diff panes now pass the pane width to delta's `-w` on every render, keeping it in sync across resizes instead of a fixed 80 columns (delta can't auto-detect a piped stdout's width). Matters most for a `side-by-side` delta setup, where it corrects real column-wrap width; plain delta output only changes a decorative rule's width. → [configuration](docs/configuration.md)
+
 ## [1.13.0] - 2026-07-16
 
 ### Added
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -111,11 +111,17 @@ instead get the file **content on stdin** and your value **replaces** the whole 
 (flags aren't merged), so a custom renderer must read stdin (glow and bat need a trailing `-`)
 and set its own flags — the token `{name}` is substituted with the file name.
 
-**Known limitation:** the full-file-diff view derives its line-numbered gutter from the `diff`
-command by appending delta's `--line-numbers` flag; if you point `diff` at a tool that rejects
-that flag (nonzero exit or spawn failure), the full-file-diff view falls all the way back to
-plain, unrendered diff text (with a notice), not just a missing gutter. Renderer timeouts and
-other limits aren't configurable.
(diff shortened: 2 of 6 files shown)


smarzban:
@trsxxii Sorry, #111 landed the same fix a few hours after you opened this, so I'm closing this one. Same root cause, same conclusion about the piped stdout.

PR #116, opened 2026-07-20
feat: launch open target (--open / HERDR_FILE_VIEWER_OPEN)

### LINKED ISSUE #109: Launch argument to open directly to a file (and optionally a line)
## Problem / motivation

I maintain a small companion herdr plugin that resolves a token (a clipboard path, a GitHub
blob URL, an agent-supplied path) to a repo-relative file, then hands it to
herdr-file-viewer so the user sees it without leaving their terminal session.

There's no way to tell the viewer "open at this file" from the outside today, so the plugin
has to drive the running viewer's own keys over the herdr socket: focus the pane, send `f` to
open the fuzzy-finder, type the repo-relative path, send `Enter`, then (for a line target) send
`:`, the line number, and `Enter` again. This works, but it's UI-scripting: it silently breaks
if the keymap changes (`f` is user-remappable per docs/configuration.md), it needs a fixed sleep
before the first keystroke so the initial tree walk doesn't eat it, and it can't distinguish "no
such file" from "the fuzzy-finder just didn't match."

## Proposed solution

A launch-time way to open straight to a file, analogous to how many editors accept `path:line`
on argv. Two shapes that would both work for my use case, in order of preference:

1. A CLI argument on the viewer binary itself, e.g.
   `herdr-file-viewer --open <path>[:line]`, where `<path>` is resolved the same way the tree
   root already resolves paths (repo-relative from the worktree root). Since the manifest's
   `[[panes]] command` is a literal argv array, a companion plugin's launcher script could just
   append `--open <path>` when it already knows the target, no socket-scripting involved.
2. Alternatively, an env var read at startup (`HERDR_FILE_VIEWER_OPEN=<path>[:line]`), if that
   fits the existing config/env conventions better than a flag , the viewer already reads
   `editor`/`open`/`reveal` overrides from config, so a same-shape env override would be
   consistent.

Either way, the target format I'd consume is the same `path[:line]` shape the viewer already
copies out via `L` (e.g. `src/app.rs:42`), so round-tripping a copied reference back in would
also just work.

## Alternatives considered

- Current workaround: driving `f` / send-text / `Enter` / `:` / line / `Enter` over the socket,
  as described above. Works but is fragile against keymap changes and needs a startup-race sleep.
- A dedicated socket verb (something like `plugin action invoke open-file-viewer --open <path>`)
  would also solve it, but a plain CLI arg feels like the smaller, more general surface and
  matches how the binary is already launched via the manifest's `command` array.

## Additional context

This is a read-only ask: I'm not requesting anything that would let an external caller select or
open outside the tree, just start focused where a caller already knows the answer. Happy to
sanity-check a proposed flag/env shape against my use case before it ships.

smarzban:
## Summary

- Launch-time **open target**: open straight to a repo-relative file via `--open <path>[:line]` or `HERDR_FILE_VIEWER_OPEN` (flag wins), so agents/companions need not key-script the fuzzy finder ([#109](https://github.com/smarzban/herdr-file-viewer/issues/109)).
- Reuses **reveal in tree** + optional **go to line** (`pending_goto`); soft-fails with a notice if the path is missing or outside the tree root.
- Range form `path:A-B`: jump to start, status notice `Opened path:A-B`, and a **1s passive highlight** (not line-select mode). Narrow/tree-only first layout **zooms** so the file is visible.
- Docs: usage guide (companion examples + copy-paste **Teach your agent** block for `AGENTS.md`), CHANGELOG, CONTEXT glossary term, ARCHITECTURE module.

## Test plan

- [x] `cargo fmt --check && cargo clippy --all-targets -- -D warnings && cargo test`
- [x] `herdr-file-viewer --open src/main.rs:10` lands on the file and line
- [x] `HERDR_FILE_VIEWER_OPEN=src/app.rs:10-20 herdr-file-viewer` (or `plugin pane open … --env …`) shows notice + brief range highlight
- [x] Missing path shows `Could not open …` and does not crash
- [x] Narrow split auto-zooms content after first layout
- [x] Docs: `docs/usage.md` open-at + teach-your-agent section reads cleanly

Closes #109


smarzban:
## Empanel — Round 2

Findings: 5 total — 0 critical · 0 high · 0 medium · 5 low · 0 info.
_Reviewed by:_ verify · models: openai-codex:gpt-5.5

### Progress since Round 1
✅ Resolved (0)
⏳ Still blocking (0)
🆕 New / regressed (0)

### Advisory (non-blocking)
- ⚪ **[LOW]** \[test-coverage\] Deferred open-target zoom lacks integration coverage — `src/app.rs:250` · 1/1 models · _by:_ verify _(test-coverage)_
  The deferred zoom behavior depends on the event loop feeding the presenter-measured content viewport back into \`set\_content\_viewport\` and scheduling a second draw. Existing unit tests cover \`set\_content\_viewport\`, but not the app-level wiring that makes launch-open zoom happen in a real narrow frame.
  _Fix:_ Add an integration/e2e test with a narrow launch-open layout that verifies the viewport feedback triggers the follow-up zoomed draw.
- ⚪ **[LOW]** \[usability\] Explicit open target still flips show\_ignored for the session — `src/controller/mod.rs:1417` · 1/1 models · _by:_ verify _(usability)_
  After \`tree.reveal\` relaxes filters for an explicit launch target, \`apply\_open\_target\` copies \`tree.show\_ignored()\` into the controller mirror. Opening a gitignored explicit path therefore leaves ignored files visible as a session preference rather than just revealing the target once.
  _Fix:_ If one-shot reveal is desired, preserve and restore the prior \`show\_ignored\` setting after selecting the target; otherwise document the persistent filter change as part of explicit-path behavior.
- ⚪ **[LOW]** \[correctness\] Open-range flash timer starts before content renders — `src/controller/mod.rs:1448` · 1/1 models · _by:_ verify _(correctness)_
  The range flash deadline is still set from \`Instant::now()\` inside \`apply\_open\_target\`, before the async render result has necessarily landed or been drawn. A slow first render can consume most or all of the intended one-second visibility window before the highlight ever appears.
  _Fix:_ Arm the timeout when the target content is first applied/drawn, or keep a pending range and start the one-second timer after the first paint that includes it.
- ⚪ **[LOW]** \[usability\] Passive range flash still consumes the close key first — `src/controller/mod.rs:2469` · 1/1 models · _by:_ verify _(usability)_
  \`close\_or\_unzoom\` still returns immediately after clearing \`open\_range\_flash\`, so pressing \`q\`/\`Esc\` during the launch range flash dismisses that layer instead of continuing to unzoom or quit. This intentional dismiss layer remains easy to regress because only the clear behavior is not directly covered.
  _Fix:_ Add a focused controller test for \`Close\` while \`open\_range\_flash\` is active, or change the close path to clear the flash and continue with the normal close action if interception is not desired.
- ⚪ **[LOW]** \[test-coverage\] E2E still does not verify the open-range flash on screen — `tests/e2e\_open\_target.rs:14` · 1/1 models · _by:_ verify _(test-coverage)_
  The pty open-target tests cover path, line, env, and precedence, but none launch a \`path:start-end\` range or assert that the range flash is visible in the live screen path. Current coverage is limited to controller/presenter-level tests, so event-loop wiring regressions could escape.
  _Fix:_ Add an e2e/integration test that launches with an open range and observes the rendered range flash, or otherwise exercises the full app draw path for range highlighting.

smarzban:
## Empanel — Round 1

Findings: 9 total — 0 critical · 0 high · 3 medium · 5 low · 1 info.
_Reviewed by:_ holistic + lens-security + lens-tests · models: anthropic:claude-opus-4-8, deepseek:deepseek-v4-pro, ollama:glm-5.2, openai-codex:gpt-5.5

### Advisory (non-blocking)
- ⚪ **[LOW]** Event-loop deferred-zoom redraw wiring is not integration-tested — `src/app.rs:253` · 1/4 models · _by:_ lens-tests _(test-coverage)_
  \`set\_content\_viewport\` now returns a bool that the event loop feeds into \`dirty = need\_redraw\` so a narrow tree-only launch open repaints once the zoom arms. The controller unit test covers the return value in isolation, but the app.rs wiring (that a returned \`true\` re-arms \`dirty\` for a second paint) is untested — the e2e tests spawn a wide pty, so the narrow-zoom path never fires. A refactor that dropped the \`need\_redraw\` propagation would keep all tests green while leaving a narrow-pane launch stuck on a blank tree-only frame.
  _Fix:_ Add an e2e that launches \`--open\` in a deliberately narrow pty (below the two-column threshold) and asserts the opened file's content becomes visible (the zoom fired and a redraw followed).
- ⚪ **[LOW]** \[correctness\] open-range flash 1s deadline starts before content renders — `src/controller/mod.rs:1445` · 3/4 models · possible same-defect siblings — combined 4/4 · _by:_ holistic _(correctness)_
  The passive range-highlight \`deadline\` is \`Instant::now() + OPEN\_RANGE\_FLASH\` set at apply time, but the file's render is dispatched off-thread and lands later via \`poll\`. With the live renderer (up to the 5s RENDER\_TIMEOUT, or even a modest glow/delta delay), the 1-second flash can expire before the file content is ever painted, so the range highlight the feature promises may never actually be visible.
  _Fix:_ Arm the flash relative to when the target render is applied (e.g. start/refresh the deadline in \`poll\` when the awaited seq lands) rather than at apply time, so the 1s window covers the frames where the content is actually on screen.
  _Possible sibling (shared topic):_ \[test-coverage\] passive range highlighting is not rendered by any test — `src/presenter.rs:691`
- ⚪ **[LOW]** \[correctness\] Passive range highlight intercepts quit/close — `src/controller/mod.rs:2465` · 2/4 models · _by:_ holistic, lens-tests _(correctness)_
  The new passive open-range flash is documented and commented as display-only/no key capture, but \`Intent::Close\` clears it and returns before the normal close path. Because both Esc and q map to Close, pressing q shortly after opening a range only dismisses the highlight instead of quitting.
  _Fix:_ Do not consume Close for \`open\_range\_flash\`, or distinguish Esc-only dismissal from q/quit so the passive highlight does not block normal close behavior.
- ⚪ **[LOW]** \[test-coverage\] range flash duration contract is tested tautologically — `src/controller/mod.rs:94` · 1/4 models · _by:_ lens-tests _(test-coverage)_
  The only expiration test advances time by \`OPEN\_RANGE\_FLASH + 1ms\`, so it would still pass if the advertised roughly-1-second duration were changed accidentally. The new user-visible timeout constant is not locked by a boundary assertion.
  _Fix:_ Assert the boundary value directly, e.g. after arming a range flash, \`tick\_open\_range\_flash(now + 999ms)\` is false and \`tick\_open\_range\_flash(now + 1000ms/1001ms)\` expires, or expose the duration through a tested contract value.
- ⚪ **[LOW]** \[security\] Launch open target can surface gitignored secrets and unhide all ignored files for the session — `src/tree.rs:339` · 1/4 models · _by:_ holistic, lens-security _(security)_
  reveal() now relaxes show\_ignored (previously only changed\_only/hide\_hidden), and the change flips it permanently true for the whole session, not just for the one target. The open target is driven by HERDR\_FILE\_VIEWER\_OPEN / --open, which in herdr's plugin/companion/agent launch flow can come from a less-trusted source than the user. A target pointed at a gitignored secret (e.g. \`.env\`, credentials, build artifacts) both displays that file's contents (classify() only blocks out-of-root reads, not gitignored in-root files) and then unhides every gitignored file in the tree, defeating the gitignore-respecting default that normally keeps such files out of view.
  _Fix:_ Scope the relaxation to the single revealed target instead of flipping the global show\_ignored flag (e.g. a per-path allow-set the tree honors for that node only), or gate ignored-file reveal behind the same trust as an interactive user action so an env/argv-supplied target cannot silently unhide the whole ignored set.
- ⚪ **[INFO]** \[test-coverage\] No e2e test verifying range flash highlight renders on screen — `tests/e2e\_open\_target.rs:0` · 1/4 models · possible same-defect siblings — combined 4/4 · _by:_ holistic _(test-coverage)_
  The range flash feature (\`path:start-end\` → passive 1s highlight) has unit tests for controller state but no e2e test confirming the highlight actually renders in the terminal. The existing e2e tests only verify single-line opens. A range-flash e2e test would catch rendering regressions in the passive highlight path.
  _Fix:_ Add an e2e test that passes \`--open some.rs:1-3\`, verifies the notice text (\`Opened some.rs:1-3\`), and confirms the flash highlight is applied to the rendered content lines (e.g. by checking that a styled line appears differently).
  _Possible sibling (shared topic):_ \[test-coverage\] passive range highlighting is not rendered by any test — `src/presenter.rs:691`

- **[MEDIUM]** Go-to-line view-forcing (SyntaxContent override) for transformed files is untested — `src/controller/mod.rs:1424`
  _Dismissed:_ Cluster mixed two issues. show\_ignored resync fixed via apply\_open\_target\_resyncs\_show\_ignored\_mirror (mirror true + secret.log selected). SyntaxContent override remains exercised by pending\_goto path + apply\_open\_target\_queues\_goto\_line; not a merge blocker.
- **[MEDIUM]** \[test-coverage\] passive range highlighting is not rendered by any test — `src/presenter.rs:691`
  _Dismissed:_ Fixed in b25e650: tests/presenter.rs passive\_range\_highlight\_has\_no\_caret\_gutter asserts no ▶ and snapshots presenter\_passive\_range\_flash for LineSelectView { passive: true }.
- **[MEDIUM]** \[test-coverage\] open-target line jump e2e would pass if line suffix were ignored — `tests/e2e\_open\_target.rs:35`
  _Dismissed:_ Fixed in b25e650: controller::tests::apply\_open\_target\_line\_scrolls\_after\_poll applies open with :20, polls until pending\_goto clears, asserts content\_scroll \> 0 on a short viewport. E2E still shows file body; scroll is unit-proven.

smarzban:
## Orchestrator review (Empanel Round 2 — final)

### What this PR implements
Launch-time **open target** (`--open` / `HERDR_FILE_VIEWER_OPEN`) so agents and companions can open a repo path (optional line/range) without key-scripting the fuzzy finder. Includes lexical path normalize + root containment, reveal that can surface gitignored paths for explicit intent, narrow-pane deferred zoom, success notice, and a 1s passive range highlight. Docs cover usage + a copy-paste AGENTS.md block.

### Panel
- **Round 1:** 4 holistic (glm-5.2, gpt-5.5, opus-4-8, deepseek-v4-pro) + lens-tests ×2 + lens-security ×2 + deterministic scan. Full panel, no missing seats.
- **Lenses:** tests + security fired; others skipped (no API/schema/migrations/PII/infra).
- **Round 2:** single-model verify (gpt-5.5) after test-coverage fixes.

### Fixed (from Round 1 mediums)
- Presenter coverage for **passive** range flash (snapshot + no ▶)
- Open-target **line jump actually scrolls** after poll (unit)
- **`show_ignored` mirror resync** after reveal of a gitignored path

### Deferred / advisory (lows, intentional or residual)
- Flash timer starts at apply (1s product choice)
- Esc clears flash before quit (intentional stack)
- Explicit open of gitignored path leaves show_ignored on for the session (same filter-relax policy as `c`/`.`)
- No pty e2e for style-only range flash; no full event-loop zoom integration test

### Decision: ✅ Approve
Empanel verdict **`pass`** on HEAD (`b25e650`). Residual findings are low/advisory only; no blocking issues remain.

**Do not merge from this comment alone** if CI is still red; green CI + this gate is the bar.

PR #120, opened 2026-07-24
Minor key bindings configuration snippet improvement

pedrusck:
## Summary

A minor improvement to the code snippet sample for the keybindings configuration is applied. This allows Herdr to better manage the output by calling into a specific installed plugin instead of spawning a detached shell in the background (see [documentation](https://herdr.dev/docs/configuration/#custom-command-keybindings)).

## Changes

- Replace the use of the `shell` type with the `plugin_action` type
- Include a description in the two key bindings configuration example and remove comments.

## Test plan

*No testing is needed as only documentation is here being changed.*


smarzban:
Thanks @petrusck 👌🏻 I updated the command as well.

PR #128, opened 2026-07-28
fix: treat AltGr as typing so its characters reach their action

### LINKED ISSUE #123: AltGr keys are dropped, so ? and other AltGr characters do nothing
## What happened

On a Brazilian ABNT2 keyboard, `?` is typed with AltGr+W. Pressing it in the viewer does nothing.
No overlay, no notice, no error. The help overlay is unreachable.

Windows reports AltGr as Ctrl+Alt. `decode` in `src/input.rs` drops any event that carries a
modifier other than Shift, so the event is discarded one step before the binding lookup:

```rust
if key.modifiers.difference(KeyModifiers::SHIFT) != KeyModifiers::NONE {
    return None;
}
```

The event still carries the resolved character (`Char('?')`), so nothing about the keypress was
ambiguous. It was thrown away because of how the character was produced.

This is not limited to `?` or to ABNT2. Any bound character that needs AltGr on your layout is a
dead key. On German and Spanish layouts that includes `[` and `]`. The failure is silent, so there
is no way to tell a broken binding from a broken keyboard.

I expected the key to run its action, the way it does on a US layout.

## Steps to reproduce

1. Set the keyboard layout to Portuguese (Brazil ABNT2).
2. Open the viewer.
3. Press AltGr+W to type `?`.
4. Nothing happens. On a US layout, Shift+/ opens the help overlay.

## Environment

- Plugin version: 1.14.0

## Logs / screenshots

No output. The key produces no notice and no log line, which is the part that makes it hard to
diagnose.

## A note on the fix

The narrow fix is to accept Ctrl+Alt on a character key and drop those two modifiers before the
lookup. Ctrl alone, Alt alone, and Ctrl+Alt on a non-character key stay inert, so reserved terminal
combos such as Ctrl+C still pass through.

There is one trade-off worth your call. Windows cannot distinguish a real Ctrl+Alt+x chord from
AltGr+x, so that chord would also fire the bound action. A user who needs the chord back can rebind
the action off that character in `[keys]`.

I have this working and will open a PR.

Jprnp:
## Summary

Accept a character typed with AltGr instead of discarding it as a chord.

Windows reports AltGr as Ctrl+Alt. `decode` dropped any event carrying a modifier other than Shift,
so on a layout where a bound character IS an AltGr key, that key did nothing at all: `?` on a
Brazilian ABNT2 keyboard, `[` and `]` on German and Spanish ones. The event still carried the
resolved character, so nothing was ambiguous about what the user typed. It was thrown away one step
before the lookup, and silently, which leaves no way to tell a broken binding from a broken
keyboard.

The exception is narrow. Only Ctrl and Alt together, and only on `KeyCode::Char`. Ctrl alone, Alt
alone, and Ctrl+Alt on a non-character key such as `Ctrl+Alt+↑` all stay inert, so every reserved
terminal combo still passes straight through.

One trade-off, and it is yours to accept or reject. Windows cannot distinguish a real `Ctrl+Alt+x`
chord from `AltGr+x`, so that chord now fires the bound action too. A user who wants the chord left
to their terminal can rebind the action off that character in `[keys]`. I could not find a way to
separate the two on Windows; if you know one I will use it instead.

## Changes

- `src/input.rs`: `decode` treats Ctrl+Alt on a character key as typing and drops those two
  modifiers before the binding lookup.
- Two tests: AltGr characters reach their action (with and without Shift), and Ctrl alone, Alt
  alone, Super, and Ctrl+Alt on a non-character key stay inert.
- `docs/keys.md` and a `CHANGELOG.md` entry.

## Test plan

- [x] `cargo test` is green
- [x] Docs updated in this PR (CHANGELOG + `docs/keys.md`)
- [x] Verified live on Windows 11 with an ABNT2 layout: AltGr+W opens the help overlay, where
      before it did nothing

## Related

Closes #123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project are documented here. The format is based on
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html). Entries are short on purpose; follow the
 `→` links for the full detail.
 
+## [Unreleased]
+
+### Fixed
+- Keys typed with **AltGr** now reach their action instead of being dropped. Windows reports AltGr as `Ctrl+Alt`, which the dispatcher discarded as a chord, so a character that needs AltGr on your layout was a dead key — `?` (help) on a Brazilian ABNT2 keyboard, `[` / `]` on German and Spanish ones. Only `Ctrl+Alt` on a character key is accepted; `Ctrl` alone, `Alt` alone, and `Ctrl+Alt+↑` stay inert. → [keys](docs/keys.md)
+
 ## [1.14.0] - 2026-07-20
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project are documented here. The format is based on
 ## [Unreleased]
 
 ### Fixed
-- Keys typed with **AltGr** now reach their action instead of being dropped. Windows reports AltGr as `Ctrl+Alt`, which the dispatcher discarded as a chord, so a character that needs AltGr on your layout was a dead key — `?` (help) on a Brazilian ABNT2 keyboard, `[` / `]` on German and Spanish ones. Only `Ctrl+Alt` on a character key is accepted; `Ctrl` alone, `Alt` alone, and `Ctrl+Alt+↑` stay inert. → [keys](docs/keys.md)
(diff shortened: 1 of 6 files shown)


smarzban:
@Jprnp  Thanks for reporting this and for the solid diagnosis and the fix. I’ve pushed a few adjustments: the workaround is now Windows-only, uses an exact shared AltGr normalizer, keeps Linux/macOS behavior unchanged, makes AltGr+? both open and close help, and adds regression coverage for custom bindings and documentation.

smarzban (approved):

PR #129, opened 2026-07-28
fix: read config.toml on Windows

### LINKED ISSUE #124: config.toml is never read on Windows
## What happened

On Windows, `config.toml` is never read. Every key in it is ignored, with no error. I set
`tree_max_cols` and nothing changed, and I assumed for a long time that the setting did not work.

The cause is a chain, and each link is reasonable on its own:

1. `config_path` looks at `HERDR_PLUGIN_CONFIG_DIR`, then `$XDG_CONFIG_HOME`, then `$HOME`.
2. herdr sets `HERDR_PLUGIN_CONFIG_DIR` only for a pane it spawns from the manifest.
3. The Windows launchers cannot let herdr spawn the pane. `CreateProcessW` resolves the manifest's
   relative command against herdr's own directory and fails with error 3, which the scripts already
   document. So they run `pane split` plus `pane run <absolute .exe>` instead, and that pane is a
   plain PowerShell shell with no plugin environment.
4. Windows sets neither `XDG_CONFIG_HOME` nor `HOME`. It sets `USERPROFILE`, which `config_path`
   does not consult.
5. The path therefore resolves to a relative `.config/herdr-file-viewer/config.toml`, and
   `load_config` refuses to read a non-absolute path.

That last refusal is correct. A config file can set the `editor` command, so reading one from the
working directory would let any repository you open run a command of its choosing. The gap is that
nothing absolute was ever found on Windows.

I expected the file at `herdr plugin config-dir herdr-file-viewer` to be read, the way it is on
Linux and macOS.

## Steps to reproduce

1. On Windows, install the plugin and open the viewer with the `open-file-viewer-windows` action.
2. Put `tree_max_cols = 70` in the file printed by `herdr plugin config-dir herdr-file-viewer`.
3. Reopen the pane. The tree is still capped at the default 30 columns.
4. Open the `?` overlay, Settings tab. Every value is the built-in default.
5. Launch the same pane with `--env HERDR_PLUGIN_CONFIG_DIR=<that directory>` and the setting
   applies.

## Environment

- Plugin version: 1.14.0

## Logs / screenshots

Nothing is logged. A missing config file is the normal case, so the load path treats "not found"
as unremarkable, which is what makes this hard to spot.

## A note on the fix

Two parts. The Windows launchers can ask herdr for the directory with
`plugin config-dir herdr-file-viewer` and pass it on with `--env`, which fixes the herdr-managed
file. And `config_path` can fall back to `%USERPROFILE%\.config` when `HOME` and `XDG_CONFIG_HOME`
are absent, so a standalone `herdr-file-viewer.exe` can have a config file at all.

I have this working and will open a PR.

Jprnp:
## Summary

Make `config.toml` reachable on Windows. Today no key in it has any effect there, and nothing says
so.

The Windows launchers cannot let herdr spawn the pane from the manifest, because `CreateProcessW`
resolves the relative command against herdr's own directory and fails with error 3. The scripts
already document that and spawn the viewer themselves by absolute path. That pane is a plain shell,
and herdr injects `HERDR_PLUGIN_CONFIG_DIR` only into panes it spawns. The viewer then fell through
to `$XDG_CONFIG_HOME` and `$HOME`, neither of which Windows sets, resolved a relative path, and
`load_config` refused to read it.

That refusal is right and I have not touched it. A config file can set the `editor` command, so
reading one relative to the working directory would let any repository you open supply a command.
The gap was that nothing absolute was ever found.

## Changes

- `scripts/open-file-viewer.ps1` and `scripts/open-file-viewer-tab.ps1`: ask herdr for the directory
  with `plugin config-dir herdr-file-viewer` and pass it to the new pane or tab with `--env`. A
  failure returns an empty string and the launch proceeds on built-in defaults, so this cannot stop
  the viewer from opening.
- `src/config.rs`: `config_path` falls back to `%USERPROFILE%\.config` when `HOME` and
  `XDG_CONFIG_HOME` are both absent, so a standalone `herdr-file-viewer.exe` can have a config file
  at all. `HOME` still wins where both are set, since a shell that sets it chose to.
- Two resolver tests for the new arm and its precedence, and a content guard in
  `tests/launcher_content.rs` so the launchers cannot silently lose the variable again. That file is
  deliberately not Windows-gated, so it runs on the required matrix.
- `docs/configuration.md` and a `CHANGELOG.md` entry.

## Test plan

- [x] `cargo test` is green
- [x] Docs updated in this PR (CHANGELOG + `docs/configuration.md`)
- [x] Verified live on Windows 11: opened the pane through the `open-file-viewer-windows` action
      with no environment set by hand, and a `tree_max_cols` value in the herdr-managed config.toml
      took effect, with the `?` overlay's Settings tab showing it. Before the change the same file
      was ignored and Settings showed only defaults.

## Related

Closes #124


smarzban:
Thanks @Jprnp 🙌🏻

PR #130, opened 2026-07-28
fix: indent a tree row by its depth alone

### LINKED ISSUE #125: A file row is indented two columns less than a directory at the same depth
## What happened

In the tree, the column a name starts in is not a function of its depth. A directory row is
`indent + "▾ "`, and the arrow is two columns wide. A file row has no glyph at all, so a file is
drawn two columns left of a directory at the same depth.

Two things follow, both visible in the repo's own snapshot
(`tests/snapshots/presenter__presenter_wide.snap`):

```
│  ▾ src          dir, depth 0, name starts at column 4
│M   main.rs      file, depth 1, name starts at column 4   same column as its parent
│    ▾ inner      dir, depth 1, name starts at column 6
│A     added.rs   file, depth 2, name starts at column 6   same column as its parent
│  README.md      file, depth 0, name starts at column 2   two left of its sibling "src"
```

A file sits in exactly the column of its parent directory's name, so it reads as a sibling of its
own parent. And a top-level file sits two columns left of a top-level directory, so two siblings do
not line up.

I expected indentation to track depth: siblings in one column, a child one level in from its parent.

## Steps to reproduce

1. Open the viewer on any repository with a file inside a directory.
2. Expand the directory.
3. The child file's name starts in the same column as the directory's own name.

Or read `tests/snapshots/presenter__presenter_wide.snap`, which records the behaviour already.

## Environment

- Plugin version: 1.14.0

## Logs / screenshots

The snapshot above is the clearest record. It is not Windows-specific and does not depend on the
renderers.

## A note on the fix

Give a file row two blanks where a directory has its arrow. That makes the starting column a
function of depth alone. It shifts every file row two columns right, so the presenter snapshots
change; that shift is the change, not a side effect.

If the current alignment is deliberate I would rather hear that than send a patch, so tell me and I
will close this. I have the fix working either way.

Jprnp:
## Summary

Indent a tree row by its depth alone.

A file row reserved no space for the expand arrow, which is two columns wide, so the column a name
started in was not a function of its depth. A file at depth `d+1` began in exactly the column of its
parent directory's name at depth `d`, and a top-level file began two columns left of the directory
sitting beside it. The tree read as if a file were a sibling of its own parent.

The repo's own snapshot records it:

```
│  ▾ src          dir, depth 0, column 4
│M   main.rs      file, depth 1, column 4   same as its parent
│    ▾ inner      dir, depth 1, column 6
│A     added.rs   file, depth 2, column 6   same as its parent
│  README.md      file, depth 0, column 2   two left of its sibling "src"
```

After: `src` and `README.md` share column 4, `main.rs` and `inner` share column 6, `added.rs` is at
8.

## Changes

- `src/presenter.rs`: a file's glyph is two blanks instead of an empty string.
- Presenter snapshots updated. Every file row shifts two columns right; that shift is the change
  itself, not a side effect, so the diff is large but mechanical.
- `tests/presenter.rs`: the annotation-marker test measured the git and `@` cells at a fixed offset
  from the name, which assumed the name followed the marker immediately. It now accounts for the
  reserved glyph width.
- A `CHANGELOG.md` entry.

If the old alignment was deliberate, say so and I will close this and the issue. I could not find a
rationale in the code or docs, and the top-level case (a file two columns left of its sibling
directory) reads as unintended either way.

## Test plan

- [x] `cargo test` is green
- [x] Docs updated in this PR (CHANGELOG; no `docs/` page states the row geometry)
- [x] Read back live in a deeply nested repository: children now sit one level in from their parent
      and siblings share a column

## Related

Closes #125


smarzban:
Thanks @Jprnp  nice catch, will add it to the next release.

PR #131, opened 2026-07-28
feat: jump to the next / previous changed file (] / [)

### LINKED ISSUE #126: Jump to the next / previous changed file from the tree
## Problem / motivation

I use the viewer to review branches. A review is a walk over the changed files, one after the next.
The tree has no key for that walk.

My repos are Java, so every file sits under `src/main/java/br/com/<org>/<module>/`. With the
changed-only filter on (`c`), the tree still draws one row per path segment, so a single changed
file costs about ten rows. To reach the next changed file I press `j` ten to fifteen times through
directory rows I do not want to read.

`f` covers the case where I already know the file name. It does not help me step through a set I
have not seen yet.

`n` and `N` already do this inside a file, over search matches. The tree has no equivalent over
changed files.

## Proposed solution

Two keys that move the tree cursor to the next and the previous changed file, and skip every
directory row. Wrap at the ends with a notice, the same way `n` and `N` wrap.

The set to walk is the one the tree already filters by: the working-tree status while `d` is on,
and the baseline-aware set behind `c` and `b` otherwise. Path order keeps the walk in the same
order that `c` lists the files.

In the full tree the next changed file can sit inside a collapsed directory. The jump can expand
its ancestors to reach it, which is what `reveal` already does for the finder.

## Alternatives considered

- `c` and then `j`. This is what I do now. It costs one press per directory row.
- `f` and type the name. Good when I know the name. No use for stepping through a set.
- A flat list of changed files instead of a tree. A much bigger change, and it drops the directory
  context that the tree gives.

## Additional context

This fits the read-only posture. The keys move the cursor and the expansion state. No file and no
git state changes.

I have it working on a branch and will open a PR.

Jprnp:
## Summary

Add `]` and `[`: move the tree cursor to the next and the previous changed file. This is `n` / `N`,
but over the tree instead of search matches.

Reviewing a branch is a walk over the changed files, and the tree had no key for that walk. In a
deeply nested repository the next changed file is a long run of `j` presses past directory rows, and
under `c` the tree still draws one row per path segment.

The set walked is the one the tree is filtered by: the working-tree status while status mode (`d`)
is on, else the baseline-aware set behind `c` and `b`. Traversal follows the changed set's own path
order, so it matches the order `c` lists the files in.

Candidates come from the changed set rather than the visible rows, so the keys also work in the
full, collapsed tree: a target with no visible row is revealed, which expands its ancestors. A
candidate that can be neither found nor revealed is skipped, since a deleted file has no node
outside changed-only mode, so the key never lands on nothing. Inert without git. An empty changed
set gets a notice rather than silence.

Read-only throughout: cursor and expansion state only (AC-N1, AC-N3).

## Changes

- `src/intent.rs`: `NextChanged` and `PrevChanged`. `Intent::ALL` is now 39.
- `src/input.rs`: `next_changed` and `prev_changed` registry rows, bound to `]` and `[`, under
  "Search & jump".
- `src/tree.rs`: `TreeModel::select_changed(forward, changed) -> Option<bool>` moves the cursor and
  reports whether the move wrapped.
- `src/controller/mod.rs`: `navigate_changed`, wired to both intents.
- Extracted `resync_filter_mirrors`. The filter-mirror resync that follows a `reveal` was already
  duplicated between the finder confirm and the launch open target, and this is a third call site.
- `tests/changed_jump.rs` (7 tests over the tree model) and 5 controller tests, covering wrapping,
  the collapsed-directory case, a skipped deleted file, status mode, and the no-git and
  nothing-changed paths.
- `docs/keys.md`, `docs/configuration.md`, `docs/usage.md`, a `CHANGELOG.md` entry, and the action
  count in `config.example.toml` and `AGENTS.md`. That count read 33 and was already stale at 37; it
  is now 39.

One note on the default keys. `[` and `]` need AltGr on German and Spanish layouts, and #123 means
AltGr characters are currently dropped. That is fixed separately in #123; without it these two keys
are unreachable on those layouts. Happy to pick different defaults if you would rather not depend on
that.

## Test plan

- [x] `cargo test` is green
- [x] Docs updated in this PR (CHANGELOG + `docs/keys.md`, `docs/configuration.md`, `docs/usage.md`)
- [x] Exercised live on a Java monorepo with 15 changed files across 5 modules, 13 to 15 levels
      deep: stepping through all of them under `c`, wrapping at both ends, and reaching a changed
      file inside a collapsed module from the full tree

## Related

Closes #126


smarzban:
Thanks @Jprnp, solid feature. Review caught an ordering bug and a few reveal-fallback issues, so I pushed fixes to your branch with tests, 🙏

PR #132, opened 2026-07-28
feat: compact single-child directory chains (compact_dirs)

### LINKED ISSUE #127: Draw a chain of single-child directories as one row
## Problem / motivation

My repos are Java. Every source file sits under `src/main/java/br/com/<org>/<module>/`. In the tree
that path is about ten rows, and each row indents two columns further than the row above it. By the
time the file name appears, the indentation has taken most of the column and the name is cut off.

The changed-only view (`c`) is the worst case, because it is almost all path. Ten rows of directory
for one file I want to read.

`H` and `L` scroll the tree sideways and `<` / `>` widen the column. Both trade one problem for
another. I can read the file name, or I can see where it sits, but not both at once.

## Proposed solution

Draw a chain of single-child directories as one row, the way the "compact folders" setting works in
VS Code. `src/main/java/br/com` becomes one row instead of six, and the file under it starts one
level in rather than six.

A chain would extend only while a directory holds one subdirectory and nothing else. It would stop
at a directory that holds a file or a second entry. The row leads into the deepest directory of the
chain, so expand, collapse and the status color act on that one.

A config key, off by default. The trade depends on the repo. On a deep layout it wins outright. On
a shallow one it costs the "one row is one directory" reading of the tree and gains little.

## Alternatives considered

- `tree_max_cols` and `tree_width`. They give the tree more columns. The indentation still grows by
  one level per segment, so a deep path runs off the edge again.
- `H` / `L` horizontal scroll. It works, but it pushes the tree structure out of view while I read
  the name.
- Show the full path on each file row instead. That is more text per row, not less.

## Additional context

This fits the read-only posture. Compaction changes how rows are grouped and labelled. It does not
change which files the tree shows, and it touches no file and no git state.

I have it working on a branch and will open a PR.

Jprnp:
## Summary

Add `compact_dirs`: draw a chain of single-child directories as one row.

On a deeply nested layout the tree spends most of a narrow column on indentation. A Java or Maven
repository puts every file under `src/main/java/br/com/<org>/...`, which is one row per segment and
two more columns of indent each time, so the file names are what gets truncated. Under `c` it is
worse, because a changed-set tree is almost all path.

With the key on, that becomes one row: `src/main/java/br/com`. The row's `path` is the deepest
directory of the chain, so expand, collapse, status colour and the child walk all act on the
directory the row leads into. `Node::label` carries the joined display name and is `None` on every
ordinary row, so an uncompacted tree renders exactly as before.

A chain extends only while a directory's sole visible entry is one subdirectory. It walks the same
`entries()` the tree draws from, so the fold follows the active gitignore and hidden filters and
never merges past something the user can see. It stops at a directory holding a file, a second
entry, or nothing. The changed-only fold mirrors that over the synthesized set and stops where a
changed file lives, so a row can never lead past a file under review.

Off by default. Compaction is a real trade: far less indentation, but a row no longer maps one to
one onto a directory, and which side wins depends on how deep the repository is. It is a startup
setting, seeded once from config and carried across a re-root like the other display preferences,
since it changes the tree's shape rather than what it shows.

Folding also removes the indentation that used to signal depth, so a folded row draws its leading
segments dim and its last segment at full weight. The eye lands on the directory the row leads into,
with the path it came through as context. Only a labelled row splits, so every other row is still a
single span and the row width is unchanged, which the horizontal-scroll clamp reads.

## Changes

- `src/tree.rs`: `Node::label`, a `compact_dirs` flag with its setter, and the fold for both the
  full tree (`compact_chain`) and the changed-only tree (`compact_synthetic_chain`).
- `src/presenter.rs`: `node_name` prefers the label; `tree_row` splits a labelled row into a dim
  prefix and a full-weight last segment.
- `src/config.rs`, `src/app.rs`, `src/controller/mod.rs`: the `compact_dirs` key, resolved into
  `EffectiveSettings` and applied by `Controller::apply_compact_dirs`, carried across a re-root.
- `src/help.rs`: the effective value in the `?` overlay's Settings tab.
- `tests/compact_dirs.rs` (8 tests): the default shape unchanged, the fold and its deepest-directory
  path, expanding a folded row, the three ways a chain ends, folding past a gitignored sibling, and
  the changed-only cases. Plus a presenter render test, a config resolver test, and a controller
  test for the config wiring and its re-root carry.
- `docs/configuration.md`, `docs/usage.md`, `config.example.toml`, and a `CHANGELOG.md` entry.

## Test plan

- [x] `cargo test` is green
- [x] Docs updated in this PR (CHANGELOG + `docs/configuration.md`, `docs/usage.md`,
      `config.example.toml`)
- [x] Read back live on a Java monorepo, both on and off, in the full tree and under `c`: a
      13-segment path folds to one row and the file names fit where they did not before

## Related

Closes #127


smarzban:
Thanks @Jprnp, nice feature and a clean fold design.  I pushed a rework to your branch: foldability is now a two-entry probe with no listing cache 🙏

PR #133, opened 2026-07-30
Add page-wise scrolling (Space / PageDown / PageUp)

wasuregusa18:
### Problem

There's no page-sized movement in the viewer. `nav_down` moves the tree cursor — or the content pane — by a single row, and `tree_scroll_left`/`right` are horizontal, so reading a long file means holding `j`.

`Space`, `PageUp` and `PageDown` are already listed in the documented bindable-key surface and `parse_key_spec` already accepts them, but no intent claims them — so they can't be remapped into paging either. `scroll_lines` is mouse-wheel only.

### What this does

Adds `page_up` / `page_down` intents that move by one screenful.

They reuse the existing `navigate()` seam, so the content pane pages when focused and the tree cursor pages otherwise, and `scroll_content()`'s existing clamp keeps the offset in range — no new scrolling logic.

The step is the live `content_height` rather than a fixed stride, so paging follows a resize. `page_step()` floors it at 1, because startup runs before the first layout pass while `content_height` is still 0 — otherwise a page key would be a silent no-op there.

### Key choice

`Space` for page-down follows the pager convention (`less`, `more`, `man`, and so `bat`). I deliberately did **not** take `less`'s page-up `b`, since that's `toggle_baseline` here — `PageUp` is the default instead, and `page_up` is remappable for anyone who wants `less` parity:

```toml
[keys]
page_up = ["b", "PageUp"]
```

Both are read-only movement, so they land in the non-mutating arm of the AC-N3 exhaustive-match guard in `intent.rs`.

### Touchpoints

Followed the checklist in `AGENTS.md`: `Intent` variant + `Intent::ALL`, a `REGISTRY` row each (category `Navigation`), the controller handler, and the docs in the same PR — `docs/keys.md`, the remappable-action table in `docs/configuration.md`, and a `CHANGELOG.md` entry under `Unreleased`. Also updated `all_length_is_37` → `39` and the test `BINDINGS` completeness fixture.

Added `page_step_is_one_viewport_and_never_zero` covering the viewport-tracking and the zero-height floor.

### Verification

`cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings` are clean. `cargo test` is 754 passed / 1 failed.

That one failure is **pre-existing and not from this change** — `e2e_keyboard::search_routes_keys_to_query_and_n_cycles_and_esc_restores` fails identically on unmodified `main` in my environment (macOS arm64, Ghostty), timing out at the final `q`/Eof step. I confirmed by stashing the patch and re-running. Happy to dig in separately if it's not a known local-environment quirk.


smarzban:
Thanks @wasuregusa18, great feature. Review caught one issue: in the narrow tree-only layout the page keys moved a single row, since the content viewport is (0,0) there. I pushed a fix to your branch that pages by the focused pane's height, thanks again🙏

wasuregusa18:
> Thanks @wasuregusa18, great feature. Review caught one issue: in the narrow tree-only layout the page keys moved a single row, since the content viewport is (0,0) there. I pushed a fix to your branch that pages by the focused pane's height, thanks again🙏

@smarzban good catch, I must've missed that. Thanks for the awesome plugin.

PR #134, opened 2026-07-30
feat: configure ignored-file visibility at startup

### LINKED ISSUE #119: Add a config option to show gitignored entries by default
## Problem / motivation

In repositories where gitignored paths are still useful to inspect (for example generated output, local configuration, caches, or vendored files), the viewer always starts with ignored files and directories hidden.

The `i` key reveals them for the current session, but this must be repeated every time the viewer is opened. It can also make the initial tree look incomplete to users who expect a general-purpose file browser rather than a tracked-files-focused view.

## Proposed solution

Add a configuration option that controls the initial visibility of gitignored entries, for example:

```toml
show_ignored = true  # default: false, preserving current behavior
```

When enabled, the tree would start in the same state as after pressing `i` once. The existing `i` key should continue to toggle visibility during the session.

Suggested details:

- Keep the default `false` for backward compatibility.
- Apply it to entries ignored by `.gitignore` and Git excludes.
- Keep `.git/` always hidden.
- Document it in `config.example.toml` and `docs/configuration.md`.
- Show the effective value in the `?` help overlay's Settings section.

The name `show_ignored` would also match the existing internal tree state, though `show_gitignored` would be another clear option.

## Alternatives considered

- Press `i` after every launch. This works, but the choice is not persisted.
- Use an explicit `--open` target for one ignored file. That can reveal a known target, but it does not make the full ignored tree visible by default.
- A generalized startup filter enum could cover more modes, but a focused boolean seems sufficient for this use case and matches the existing `hide_dotfiles` startup setting.

## Additional context

Observed with `herdr-file-viewer` v1.14.0. This is a display-only preference and preserves the viewer's read-only design.

smarzban:
## Summary

Adds an opt-in `show_ignored` configuration key for users who routinely inspect generated files, local configuration, caches, or other Git-ignored paths.

- Defaults to `false`, preserving existing startup behavior.
- `show_ignored = true` starts the tree in the same state as pressing `i` once.
- The existing `i` key continues toggling ignored entries during the session.
- Repository-local Git excludes are included, while `.git/` remains hidden.
- The effective value appears in the `?` Settings section.
- Configuration, usage, example config, and changelog documentation are included.

Ignored paths can contain local credentials or generated secrets, so the docs recommend leaving the opt-in disabled unless it is needed.

Closes #119

## Verification

Focused coverage proves:

- parsing and the default-off/config-on resolution path,
- startup tree and controller state stay synchronized,
- the first interactive `i` press reverses a configured-on startup state,
- `.git/info/exclude` entries become visible while `.git/` never does,
- the effective value is displayed in help,
- both configuration references remain covered by documentation drift guards.

Checks run:

- `cargo fmt --check`: pass
- `cargo clippy --all-targets -- -D warnings`: pass
- `cargo build --release`: pass
- `cargo audit --no-yanked`: pass, with the repository's existing allowed `RUSTSEC-2026-0190` warning
- `cargo test --no-fail-fast`: every target passed except `search_routes_keys_to_query_and_n_cycles_and_esc_restores`, an unrelated macOS PTY timeout reproduced unchanged on `origin/main` at `94442c3`; the same main commit's CI matrix is green

No dependency, manifest, persistence, filesystem-write, or Git-write behavior changes.


smarzban:
## Empanel — Round 1

Findings: 4 total — 0 critical · 0 high · 0 medium · 3 low · 1 info.
_Reviewed by:_ holistic + lens-security + lens-spec · models: opencode-go:deepseek-v4-pro, ollama:glm-5.2, anthropic:claude-opus-5, openai-codex:gpt-5.6-terra

### Advisory (non-blocking)
- ⚪ **[LOW]** \[docs\] Over-long line in the tree section breaks the file's wrap convention — `docs/usage.md:28` · 1/4 models · _by:_ holistic _(maintainability)_
  The inserted sentence leaves \`setting off unless you need it; \`.git/\` itself always stays hidden. The tree's \*\*top border names the root\*\* directory and its\` at \~140 columns, while every other line in \`docs/usage.md\` wraps at \~100; the paragraph now reflows inconsistently in diffs and editors.
  _Fix:_ Re-wrap the paragraph at the file's \~100-column width so the new sentence does not leave one very long line.
- ⚪ **[LOW]** \[test-coverage\] No test pins app::run's startup wiring of apply\_show\_ignored — `src/app.rs:122` · 3/4 models · _by:_ holistic, lens-spec _(test-coverage)_
  The spec criterion "true starts in the same state as pressing i once" depends on app::run calling controller.apply\_show\_ignored(eff.show\_ignored) at startup. Controller-level tests prove apply\_show\_ignored(true) puts the tree in the right state, and config tests prove show\_ignored resolves to true, but no test asserts that run() actually invokes apply\_show\_ignored with the resolved value (or that it is called before the first draw). Dropping that single line would leave every test green while completely breaking the feature at actual startup — a config \`show\_ignored = true\` would never reach the tree. This mirrors the pre-existing apply\_hide\_dotfiles seam, which has the same unpinned integration gap.
  _Fix:_ Add a test that exercises the app::run startup path (or a shared wiring helper extracted from it) with a config of show\_ignored=true and asserts the resulting controller's tree shows an ignored file at launch, so removing the apply\_show\_ignored call fails a test.
- ⚪ **[LOW]** \[test-robustness\] New exclude test assumes \`.git/info/\` exists after \`git init\` — `tests/tree\_filters.rs:46` · 1/4 models · _by:_ holistic _(test-coverage)_
  \`fs::write(dir.path().join(".git/info/exclude"), ...).unwrap()\` relies on git's default template having created \`.git/info/\`; a CI/dev environment with \`init.templateDir\` pointed at a custom or empty template creates no \`info/\` dir, and the write then panics with a confusing \`NotFound\` rather than a filter assertion failure.
  _Fix:_ Create the directory first: \`fs::create\_dir\_all(dir.path().join(".git/info")).unwrap();\` before writing \`exclude\`.
- ⚪ **[INFO]** \[docs\] \`show\_ignored = true\` does not widen the \`f\` finder index — `docs/configuration.md:69` · 1/4 models · _by:_ holistic _(correctness)_
  The tree reveals ignored entries but \`index::build\` still walks with \`git\_ignore(true)\`, so the fuzzy finder never lists them. This exactly matches pressing \`i\` (which is the stated contract), but the one-line doc "true to show gitignored files at startup" can be read as widening search scope too.
  _Fix:_ If worth clarifying, add "(the \`f\` finder index stays gitignore-aware)" to the configuration.md line or the usage.md sentence.

smarzban:
## Empanel orchestrator review, Round 1

This PR adds the opt-in `show_ignored` startup setting without changing the default or introducing a second filtering path. The resolved config value is applied before launch-target handling, updates the controller mirror and existing tree filter together, preserves runtime `i` toggling, and leaves `.git/` unconditionally hidden. Repository-local Git excludes, help output, reference docs, the example config, usage guidance, and the changelog are covered.

### Verification

- Empanel deterministic scan: clean
- Panel: 12/12 votes across four model families
- Holistic: four seats
- Security lens: four seats, fired because local config can expose ignored paths containing secrets
- Spec lens: four seats, fired against issue #119's contract
- CI: all seven jobs pass, including both macOS and Ubuntu toolchains, cargo-audit, lint, and advisory Windows
- No finding was dismissed or overridden

### Lens decisions

- Tests: skipped because focused Rust tests were added in four test surfaces; Empanel's suggestion did not recognize Rust integration-test naming.
- Spec: fired because issue #119 defines explicit behavior and documentation requirements.
- Security: fired because config parsing and initial visibility of potentially sensitive ignored files touch a trust boundary.
- Privacy: skipped, no personal or sensitive data is stored, logged, or transmitted.
- Contracts: skipped, no public HTTP or asynchronous message contract changed.
- Migrations: skipped, no persistent-state migration exists.
- Subtle correctness: skipped, no concurrency, cache, date, or time behavior changed.
- Simplify: skipped, the implementation reuses the existing filter with one scalar setting and no new abstraction.
- Performance: skipped, the opt-in startup path invokes the existing `i` filter behavior once and adds no new algorithm.
- Infrastructure: skipped, no CI or deployment surface changed.

### Advisory, not merge-blocking

Empanel reported three lows and one informational note: the new usage paragraph has one long Markdown line; `app::run`'s one-line config-to-controller call is inspected but not pinned by an end-to-end startup test; the Git-exclude test assumes Git created `.git/info/`; and the finder remains gitignore-aware even when the tree starts with ignored entries shown. I checked each against the code. They are real documentation or test-hardening opportunities, but none changes runtime correctness, security, or the accepted contract, and all CI jobs pass.

### Decision

✅ **Approve.** No blocking finding is deferred. The PR is ready for the maintainer's morning review at head `b2f29befcd998460b2dac5d605784e5bc06ba720`.

PR #136, opened 2026-08-01
feat: add release details and project spotlight

smarzban:
## Status

Ready for review. This adds release details and one project spotlight to the existing daily update check.

## What changed

- `?` opens Help on **What's New**, ordered as available release details and install guidance, project spotlight, then embedded release history.
- One advisory status row covers both notice types. `u` dismisses that row for the current session while What's New remains readable.
- `update_check` controls both notice types. There is no separate spotlight or persistent dismissal setting.
- Release discovery uses isolated system Git, document retrieval uses optional system curl, and user proxy and CA trust remain inherited.
- Retrieval is fixed-source, bounded, fail-silent, and covered by one 15-second deadline. Remote content is display-only and terminal-neutralized.
- The safe-to-delete cache publishes one complete atomic snapshot with last-writer-wins semantics.

## Verification

- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test` (492 library tests plus all integration binaries)
- Forced remote-notice PTY safety test
- `cargo insta test --accept --test presenter` (121 passed, no snapshot changes)
- `cargo build --release`
- `cargo audit`
- Agent SDLC consistency check: 0 findings

## Review ask

Please review the bounded source lifecycle, session-only status dismissal, and release-first What's New behavior.


smarzban:
## Portable merge gate: pass ✅

**Status:** the full 49-file diff was reviewed at `2776da9884e0b600a40b3665216669c10e9db977` against AC-1 through AC-58, NC-1 through NC-11, the design, trust boundaries, tracked docs, and production code. No blocking findings were found.

Empanel was unavailable in this harness, so ship used its portable whole-PR reviewer fallback.

### Findings

- Critical: 0
- High: 0
- Medium: 2, both adjudicated nonblocking
- Low: 4
- Info: 4

Medium advisories:

1. Cached exact-release details are not re-filtered if the running version later advances to an intermediate version below the detected release. This can duplicate an already-installed changelog section in What's New, without changing the detected update or status row. Relevant criterion: AC-14.
2. User docs call the remote portion “Available updates,” but the composer does not render that literal heading. The content and install command are present in the specified order, but the named label is not visible. Relevant criterion: AC-56.

Low advisories cover a missing test pin for the 15-second constant, the stricter 256 KiB discovery cap versus AC-8's 1 MiB test wording, possible drift between the fixed repository literal and `CARGO_PKG_REPOSITORY`, and empty-cache publication after a transient cache read error.

The review verified the fixed-authority credential-free gateway, shared deadline, independent fail-silent outcomes, display-only neutralization, dismissal scoping, cross-process cache serialization, shutdown behavior, unchanged input/config surfaces, and hermetic tests.

**Review ask:** please decide whether to merge with these advisories or address any of them in a follow-up.

smarzban:
## Post-open fix review: pass ✅

**Status:** reviewed at `1fcc310239e96939e559c7799e059122ebd54bf7`, with 0 Critical and 0 Important findings.

Resolved:

- Re-filter cached release details after an intermediate viewer upgrade.
- Fix the Windows Unix-only test helper warning.
- Reduce cache-lock retry work so macOS shutdown stays within the existing bound.
- Remove the standalone remote-notices document and trim feature documentation to concise key, config, usage, security, and architecture notes.
- Remove the incorrect visible “Available updates” heading claim.

Snapshots remain unchanged. Local fmt, clippy, full tests, forced PTY safety test, release build, and audit all pass.

smarzban:
## Empanel — Round 1

Findings: 14 total — 0 critical · 0 high · 5 medium · 8 low · 1 info.
_Reviewed by:_ holistic + lens-security + lens-subtle-correctness · models: opencode-go:deepseek-v4-pro, ollama:glm-5.2, anthropic:claude-opus-5, openai-codex:gpt-5.6-terra
⚠️ **Coverage:** 12/13 planned reviewer passes voted — lost: lens-spec (spec chain is local-only in this repo; conformance not checked)
✅ **Dismissal audit:** all 1 dismissal justification(s) held against the code. (1 advisory-calibration dismissal(s) exempt from the audit requirement.)
⚠️ **Dismissal audit:** 1 entry matched no dismissed finding and had NO effect (cc1eade9be330a2b5c610e5472bedca846ae1e9b1bb351e215c540f87eb19b61) — if a dismissal was meant to be covered, regenerate the audit from this round's clusters.json.

### Must fix
- 🟠 **[MEDIUM]** \[performance\] Bound synchronous fallback conversion for remote documents — `src/update/compose.rs:105` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(performance)_
  Every document is converted with \`to\_text\` before the first deadline check, on the input thread. A valid 1 MiB remote spotlight or changelog made of very short lines can expand into hundreds of thousands of ratatui lines/spans, causing large allocation and an unbounded Help-open pause or memory spike despite the advertised 200 ms budget.
  _Fix:_ Apply a conservative line/structured-output cap when accepting remote documents (and discard or truncate documents exceeding it), or move fallback conversion off the input thread with a deadline-aware bounded representation so \`to\_text\` cannot exceed the Help-open budget.
  _Possible sibling (shared topic):_ \[test-coverage\] The replacement for the deleted HELP\_RENDER\_TIMEOUT test asserts only a constant, no longer pinning the input-thread bound — `src/controller/mod.rs:3334`
  _Possible sibling (co-located — verify same defect):_ \[performance\] What's New composition does unbounded work on the input thread outside the 200 ms budget — `src/update/compose.rs:99`
  _Possible sibling (shared topic):_ \[correctness\] Accepted spotlight title is unbounded, so remote text can push the status-row hints off screen (and is re-scanned every frame) — `src/update/spotlight\_policy.rs:66`
  _Possible sibling (shared topic):_ \[time\] Composer deadline test asserts non-zero remaining budget for all three documents and can flake under load — `tests/whats\_new\_composer.rs:387`
- 🟠 **[MEDIUM]** \[security\] Enforce the document cap while curl is downloading — `src/update/gateway.rs:232` · 2/4 models · _by:_ holistic, lens-security _(security)_
  \`--max-filesize\` only rejects responses whose size is known up front (for example via Content-Length); a chunked/no-length response can still be written in full to the temporary \`body\` file before the later \`read\_bounded\` check. A compromised source or inherited proxy can therefore consume arbitrary temporary-disk space during the 15-second transfer despite the advertised 1 MiB bound.
  _Fix:_ Stream the response through a bounded pipe reader and kill/reap curl immediately after 1 MiB plus the fixed status trailer (for example use \`--output -\` and parse the final write-out status), rather than writing an unbounded body file and checking it only after curl exits.
- 🟠 **[MEDIUM]** \[security\] Streamed document responses can bypass the 1 MiB cap and fill disk — `src/update/gateway.rs:234` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ lens-security _(security)_
  \`--max-filesize\` only rejects a response when curl learns an oversized size before transfer; chunked responses or a peer that lies/omits Content-Length are still written in full to the temporary \`body\` file. Since the file is only bounded after curl exits, a hostile remote endpoint/proxy can consume arbitrary disk space for the 15-second transfer window.
  _Fix:_ Do not write the transfer directly to an unmonitored file: stream the body through a capped reader (with separately delimited status), or actively poll/enforce the output file size and kill curl as soon as it exceeds \`DOCUMENT\_MAX\_BYTES\`.
  _Possible sibling (shared topic):_ \[correctness\] Accepted spotlight title is unbounded, so remote text can push the status-row hints off screen (and is re-scanned every frame) — `src/update/spotlight\_policy.rs:66`
  _Possible sibling (shared topic):_ \[time\] Composer deadline test asserts non-zero remaining budget for all three documents and can flake under load — `tests/whats\_new\_composer.rs:387`

### Advisory (non-blocking)
- ⚪ **[LOW]** \[caching\] Help's What's New body is not invalidated when a background snapshot replaces the notice state — `src/controller/mod.rs:3209` · 1/4 models · _by:_ lens-subtle-correctness _(caching)_
  \`poll\` replaces \`notice\_snapshot\` and redraws, but the What's New / About bodies are composed once in \`open\_help\` and stored in \`HelpState\`. If the refresh lands while Help is open (the worker can take up to 15 s, and Help is reachable immediately at startup), the status row updates to the new release/spotlight while the overlay keeps showing the pre-refresh content until the user closes and reopens it — an inconsistent view of the same snapshot.
  _Fix:_ When \`poll\` applies a new snapshot and \`Modal::Help\` is active, recompose the affected help sections (or mark them dirty so the next draw recomposes) so the overlay and the status row always project the same snapshot.
- ⚪ **[LOW]** \[test-coverage\] The replacement for the deleted HELP\_RENDER\_TIMEOUT test asserts only a constant, no longer pinning the input-thread bound — `src/controller/mod.rs:3334` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic _(test-coverage)_
  \`help\_composition\_timeout\_is\_the\_single\_200ms\_budget\` only checks \`WHATS\_NEW\_COMPOSE\_TIMEOUT == 200ms\`. The old test justified a real worst-case bound for one render; the new path renders up to three documents and adds unbounded pre/post text conversion, so nothing now guards the AC-22 budget that the deleted constant existed to protect.
  _Fix:_ Add a test that drives \`open\_help\` (or \`compose\_whats\_new\` with the production renderer adapter) against a maximal spotlight + changelog and asserts the wall-clock block stays inside the responsiveness budget.
  _Possible sibling (shared topic):_ \[performance\] Bound synchronous fallback conversion for remote documents — `src/update/compose.rs:105`
- ⚪ **[LOW]** \[caching\] Orphaned \`update-check.json.\*.tmp\` staging files accumulate unbounded when the process exits mid-publish — `src/update/cache.rs:163` · 1/4 models · _by:_ lens-subtle-correctness _(caching)_
  \`store\` runs on a detached background thread; if the viewer exits (user quits, pane closes) between \`create\_staging\_file\` and the \`rename\`, the uniquely named \`.tmp\` file is never removed — the cleanup only runs on an in-process error path. Nothing in the codebase sweeps the cache dir, so repeated quit-during-refresh sessions leave a growing pile of multi-MiB staging files in the user's cache directory.
  _Fix:_ Remove stale \`update-check.json.\*.tmp\` siblings (older than, say, an hour) at the start of \`store\` or during \`load\`, mirroring the sweep the test helpers already do for temp dirs.
- ⚪ **[LOW]** \[performance\] What's New composition does unbounded work on the input thread outside the 200 ms budget — `src/update/compose.rs:99` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic, lens-security, lens-subtle-correctness _(performance)_
  \`compose\_whats\_new\` pre-computes \`render::to\_text\` fallbacks for every document (spotlight body up to 1 MiB plus the embedded changelog) before the deadline is consulted, and \`delegate\_markdown\_section\` then runs \`to\_text\` on renderer stdout (capped at 16 MiB) after the deadline — all synchronously on the input thread. \`WHATS\_NEW\_COMPOSE\_TIMEOUT\` therefore bounds only the subprocess wait, not the actual Help-open block, which remote content size can inflate well past the AC-22 responsiveness budget.
  _Fix:_ Compute each document's fallback lazily (only on the expired/failed path) or bound the pre-neutralized/rendered size, and pin the total input-thread cost with a test that measures a large-spotlight Help open rather than only asserting the constant's value.
  _Possible sibling (co-located — verify same defect):_ \[performance\] Bound synchronous fallback conversion for remote documents — `src/update/compose.rs:105`
- ⚪ **[LOW]** \[error-handling\] Abandoned refresh leaks private probe directories and orphan child processes when the viewer quits mid-check — `src/update/mod.rs:192` · 1/4 models · _by:_ holistic _(error-handling)_
  The refresh coordinator runs on a detached thread and cleans up its \`herdr-fv-probe-\*\` dirs only after each call returns. Quitting the viewer inside the 15 s window kills the thread at process exit, leaving up to three temp dirs (each possibly holding a partially downloaded body) plus reparented \`git\`/\`curl\` children; nothing sweeps them on the next launch.
  _Fix:_ Sweep stale \`herdr-fv-probe-\*\` directories (and orphan \`update-check.json.\*.tmp\` staging files) older than a threshold at startup, so repeated quit-during-check sessions cannot accumulate temp state.
- ⚪ **[LOW]** \[time\] Refresh worker stamps the cache with the session-start clock, not the time of the successful check — `src/update/mod.rs:263` · 1/4 models · _by:_ lens-subtle-correctness _(time)_
  \`now\_unix\` is captured in \`start\_default\_with\` before the cache load and before the background thread runs, yet it is written as \`last\_check\_unix\` and as the spotlight \`retrieved\_at\_unix\` after a refresh that may consume the full 15 s \`REFRESH\_TIMEOUT\`. The persisted timestamps therefore claim the data is older than it is, and if \`SystemTime::now()\` fails the fallback \`0\` is persisted as a successful-check time, forcing a refresh on every subsequent launch forever.
  _Fix:_ Sample the wall clock inside the worker at publish time (with the same \`unwrap\_or\` guard) and use that for \`last\_check\_unix\` / \`retrieved\_at\_unix\`; keep the session-start value only for the pure \`decide\` freshness projection.
- ⚪ **[LOW]** \[correctness\] Accepted spotlight title is unbounded, so remote text can push the status-row hints off screen (and is re-scanned every frame) — `src/update/spotlight\_policy.rs:66` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ holistic, lens-security _(correctness)_
  \`accept()\` takes the entire first non-blank \`\# \` line as the status title with no length cap — up to \~1 MiB, since the document cap is the only bound. \`format\_status\` then builds \`Spotlight: {title} · ? details · u dismiss\` for a single clipped row, so a long remote title silently hides the \`details\`/\`dismiss\` hints; the presenter also re-allocates and byte-scans that whole title (\`sanitize\_control\`) on every redraw.
  _Fix:_ Truncate the neutralized title at the policy boundary (e.g. first N grapheme clusters / \~60 columns with an ellipsis) before storing it in \`AcceptedSpotlight::title\`, so the status line is genuinely one bounded row and the per-frame work is O(row width).
  _Possible sibling (shared topic):_ \[performance\] Bound synchronous fallback conversion for remote documents — `src/update/compose.rs:105`
  _Possible sibling (shared topic):_ \[security\] Streamed document responses can bypass the 1 MiB cap and fill disk — `src/update/gateway.rs:234`
- ⚪ **[LOW]** \[time\] Composer deadline test asserts non-zero remaining budget for all three documents and can flake under load — `tests/whats\_new\_composer.rs:387` · 1/4 models · possible same-defect siblings — combined 2/4 · _by:_ lens-subtle-correctness _(time)_
  The test sleeps 10 ms per document inside a 200 ms absolute budget and then asserts \`renderer.calls.len() == 3\` and \`!call.observed\_remaining.is\_zero()\`. Since \`compose\_whats\_new\` skips delegation entirely once the deadline expires, any scheduling stall (loaded CI, the \~MiB \`to\_text\` pre-pass) between \`Instant::now()\` and the third render makes the composer use the fallback instead, failing both assertions non-deterministically.
  _Fix:_ Pass an \`opened\_at\` far in the future or inject a controllable clock/deadline instead of relying on real wall-clock headroom, and assert the ordering property (equal deadlines, monotonically decreasing remaining) only over the calls that occurred.
  _Possible sibling (shared topic):_ \[performance\] Bound synchronous fallback conversion for remote documents — `src/update/compose.rs:105`
  _Possible sibling (shared topic):_ \[security\] Streamed document responses can bypass the 1 MiB cap and fill disk — `src/update/gateway.rs:234`
- ⚪ **[INFO]** \[maintainability\] New module cycle between \`help\` and \`update\`, and \`help::changelog\_display\` is now production-dead — `src/help.rs:15` · 1/4 models · _by:_ holistic _(maintainability)_
  \`help::released\_changelog\` now calls \`update::release\_policy::release\_sections\` while \`update::compose::compose\_whats\_new\` calls back into \`help::released\_changelog\`, creating a two-way dependency between a presentation module and the update feature. \`changelog\_display()\` is left public but is referenced only by its own unit test.
  _Fix:_ Move \`released\_changelog\` into \`update::release\_policy\` (or a neutral module) so the dependency points one way, and delete \`changelog\_display()\` or fold its assertions into the \`released\_changelog\` test.

- **[MEDIUM]** \[caching\] Cross-process read-modify-write on update-check.json silently discards another viewer's refresh — `src/update/mod.rs:210`
  _Dismissed:_ Verified in code: cache.rs store() documents the contract as last-complete-writer-wins for an advisory, safe-to-delete cache (unique staging file + atomic rename; readers always see a complete snapshot). This design deliberately replaced a cross-process lock/CacheWriter protocol in this PR's rework round with the owner's explicit approval (PR \#136 discussion). Worst case traced in mod.rs start\_with: a concurrent pane's refresh result or timestamp is overwritten, costing at most one redundant once-daily re-check of public advisory data; no user data and no display correctness is affected, since each session's in-memory snapshot is self-consistent.
- **[MEDIUM]** \[caching\] Concurrent refreshes can overwrite a newer cache snapshot — `src/update/mod.rs:298`
  _Dismissed:_ Same defect as claim 1b92154e (cross-process last-writer-wins on the advisory cache), raised under the co-located cluster. Same code-checked justification: the deliberate, documented, owner-approved trade-off for an advisory cache; worst case is one redundant daily refresh, not data loss visible to the user in-session.

smarzban:

**What this round did:** full 4-model panel (opus-5, gpt-5.6-terra, deepseek-v4-pro, glm-5.2) on holistic, plus `lens-security` and `lens-subtle-correctness` on all four roster seats (fired for the subprocess/untrusted-input surface and the deadline/concurrency logic respectively; `lens-spec` recorded as lost coverage — the spec chain is local-only). Deterministic scan: clean. All 12 seats voted; the two zero-finding seats were verified as genuine deep reads (100k–3M input tokens), not dead seats.

**Confirmed (blocking), verified in the code:**
1. **Chunked downloads bypass `--max-filesize`** (2 models, 4 claims): curl only enforces `--max-filesize` when the size is known up-front; a chunked/no-`Content-Length` response streams unbounded into the private body file for up to the 15 s window, and `read_bounded` only bounds what is read back *after* curl exits. On a tmpfs `$TMPDIR` that is a transient RAM-exhaustion vector, and SECURITY.md's "bounded transient body" claim currently overstates what is enforced. Real; must fix (bound the write, and/or make SECURITY.md accurate).
2. **What's New pre-neutralization runs outside the 200 ms budget** (1 medium + 3 corroborating low claims from a second model): `compose_whats_new` runs `to_text` over every document before the first deadline check, and the post-render `to_text` of up to 16 MiB renderer stdout also sits outside it — all on the input thread. Bounded by existing caps, but pathological remote content can push Help-open past the repo's load-bearing AC-22 budget. Confirmed because two models converged on it and AC-22 is constitutional in this repo.

**Dismissed (audited, all upheld by an independent model):** the cross-process last-writer-wins cache write (both claims) — that is the deliberate, documented, owner-approved design for an advisory cache, replacing the lock protocol removed in the rework; worst case is one redundant daily re-check. The 20 MiB startup-parse concern — the read is capped before parse, legitimate files are ≤ ~6 MiB, and a larger file requires the local user tampering with their own cache (out of threat model). Follow-up logged: lower `CACHE_MAX_BYTES`.

**Advisory items worth folding into the fix round:** the unbounded spotlight title, the Help-body-not-invalidated-on-refresh nit, staging/probe-dir orphans on quit-mid-refresh, the session-start timestamp stamping, and the load-sensitive composer test. The spotlight title/body size cap in particular would shrink blocking finding 2's worst case at the source.

**Decision: 🔴 Request changes** (agrees with the spine's BLOCK verdict).

smarzban:
## Empanel — Round 2

Findings: 5 total — 0 critical · 0 high · 1 medium · 3 low · 1 info.
_Reviewed by:_ holistic · models: anthropic:claude-opus-5, openai-codex:gpt-5.6-terra

### Progress since Round 1
✅ Resolved (3): \[performance\] Bound synchronous fallback conversion for remote documents; \[security\] Enforce the document cap while curl is downloading; \[security\] Streamed document responses can bypass the 1 MiB cap and fill disk — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (1): \[security\] Check the body cap before accepting completed curl output

### Carry-forward: cleared by the verifier (3)
A `resolved` disposition with evidence cleared these prior gating claims:
- \[security\] Streamed document responses can bypass the 1 MiB cap and fill disk (`src/update/gateway.rs:234`)
- \[security\] Enforce the document cap while curl is downloading (`src/update/gateway.rs:232`)
- \[performance\] Bound synchronous fallback conversion for remote documents (`src/update/compose.rs:105`)

### Must fix
- 🟠 **[MEDIUM]** \[security\] Check the body cap before accepting completed curl output — `src/update/gateway.rs:459` · 1/2 models · possible same-defect siblings — combined 2/2 · _by:_ holistic _(security)_
  If curl finishes an oversized chunked response before the first 50 ms timeout (or between polls), \`recv\_timeout\` returns \`Ok\` and this branch breaks without ever invoking \`over\_cap\`; the oversized regular file has already been written. \`read\_bounded\` later rejects it for memory purposes, but cannot uphold the documented on-disk cap and a fast response can consume an arbitrary amount of temporary disk space during that interval.
  _Fix:_ At minimum invoke \`over\_cap()\` before accepting the received status (and add a fast finite oversized-transfer regression test). To actually enforce a disk ceiling rather than merely detect it after a polling interval, route curl's body through a bounded stream/pipe that stops accepting bytes at \`DOCUMENT\_MAX\_BYTES\` instead of letting curl write directly to a regular file.
  _Possible sibling (shared topic):_ \[maintainability\] Over-cap cached release details are never cleared from the persisted cache (unlike the spotlight path) — `src/update/mod.rs:246`
  _Possible sibling (shared topic):_ \[correctness\] Spotlight title cap counts chars, not display columns — the stated goal (keep status hints on screen) is not met — `src/update/spotlight\_policy.rs:42`
  _Possible sibling (shared topic):_ \[correctness\] \`.take(MAX\_TITLE\_CHARS)\` before \`.trim()\` can withdraw a spotlight whose heading is whitespace-padded — `src/update/spotlight\_policy.rs:96`

### Advisory (non-blocking)
- ⚪ **[LOW]** \[maintainability\] Over-cap cached release details are never cleared from the persisted cache (unlike the spotlight path) — `src/update/mod.rs:246` · 1/2 models · possible same-defect siblings — combined 2/2 · _by:_ holistic _(maintainability)_
  When fetched details exceed \`MAX\_DETAILS\_BYTES\`, \`fetched\` (and therefore \`persisted\_release\_details\`) is \`None\`, so an already-persisted oversized \`release\_details\` for the same \`latest\_seen\` survives untouched and is re-filtered on every startup and re-fetched every 24h forever; the spotlight path, by contrast, clears its cache field on withdrawal.
  _Fix:_ On the \`Fetch\` branch, explicitly clear \`refresh\_cache.release\_details\` when the fetch yields no usable (in-cap) details, so an over-cap entry self-heals like the spotlight does.
  _Possible sibling (shared topic):_ \[security\] Check the body cap before accepting completed curl output — `src/update/gateway.rs:459`
- ⚪ **[LOW]** \[correctness\] Spotlight title cap counts chars, not display columns — the stated goal (keep status hints on screen) is not met — `src/update/spotlight\_policy.rs:42` · 1/2 models · possible same-defect siblings — combined 2/2 · _by:_ holistic _(correctness)_
  The doc claims the cap exists so "a hostile length must not be able to push the status row's \`details\`/\`dismiss\` hints off screen", but 160 \*chars\* (up to 320 columns for wide CJK) is far wider than a typical 80-column status row, which \`draw\_remote\_notice\_status\` renders as a single clipped \`Line\` — the hints are still pushed off screen.
  _Fix:_ Either lower the cap and/or measure in display columns (e.g. \`unicode\_width\`/ratatui \`Line::width\`), or reword the doc to say the cap merely bounds the retained string rather than guaranteeing the hints remain visible; alternatively truncate the title at format time in \`status::format\_status\` against the known row width.
  _Possible sibling (shared topic):_ \[security\] Check the body cap before accepting completed curl output — `src/update/gateway.rs:459`
- ⚪ **[LOW]** \[correctness\] \`.take(MAX\_TITLE\_CHARS)\` before \`.trim()\` can withdraw a spotlight whose heading is whitespace-padded — `src/update/spotlight\_policy.rs:96` · 1/2 models · possible same-defect siblings — combined 2/2 · _by:_ holistic _(correctness)_
  Truncation happens before trimming, so a heading with more than 160 leading spaces (e.g. \`\# \` + 200 spaces + \`Title\`) now truncates to all-whitespace, trims to empty, and returns \`None\` — a previously accepted document becomes a conclusive \`Withdrawn\`, which also throttles retrieval for 24h.
  _Fix:_ Trim the neutralized/filtered title first and take the cap afterwards (e.g. collect → trim → \`chars().take(MAX\_TITLE\_CHARS)\`), so truncation only shortens an otherwise valid title.
  _Possible sibling (shared topic):_ \[security\] Check the body cap before accepting completed curl output — `src/update/gateway.rs:459`
- ⚪ **[INFO]** \[maintainability\] Doc link from a production item to a \`\#\[cfg(all(test, unix))\]\` function is a broken intra-doc link — `src/update/gateway.rs:418` · 1/2 models · _by:_ holistic _(maintainability)_
  \`capture\_stdout\_bounded\` is now compiled only under \`cfg(test, unix)\`, but the non-test \`capture\_stdout\_bounded\_watched\` doc comment links to it, so \`cargo doc\` (and any future rustdoc lint gate) reports a broken intra-doc link; CI does not currently run \`cargo doc\`, so this is latent.
  _Fix:_ Reference the still-public \`capture\_stdout\_bounded\_with\_spawner\` instead, or drop the bracketed link and mention the name in plain prose.

smarzban:
## Orchestrator review — Round 2 (verification)

**Scope:** verifier (deepseek-v4-pro, disposition channel) over the three Round-1 gating claims + two fix-scoped reviewer seats (opus-5, gpt-5.6-terra) over `git diff 91caf8e...61bc286` + a fresh deterministic scan (clean).

**Progress:** all three Round-1 blockers are **resolved** with code-cited evidence — the curl body watchdog bounds the on-disk transfer that `--max-filesize` cannot, and the spotlight/release-details caps bound the compose pre-pass at the policy boundary.

**New this round (defects in the fix code itself, caught by the fix-scoped seats):** I am **confirming** the one medium as blocking rather than waving it through as a lone finding: when curl completes an oversized chunked body *between* watchdog polls, the capture's `Ok` path accepts the status without a final `over_cap` check — memory stays bounded and the file is deleted milliseconds later, but the invariant SECURITY.md now documents ("an over-cap transfer is killed") is not enforced on that path. Also fixing three cheap fix-regressions from the advisory list: title truncation before trim can withdraw a whitespace-padded heading, the title *line* itself is unbounded (so a 1 MiB heading still reaches the cache and the startup neutralize pass), and a broken rustdoc link to a now-test-only function. Deferred as follow-ups: char-vs-display-column title width (would need a width dep for an advisory row) and clearing over-cap cached details (unreachable state — the caps ship in the same release as the feature; no released build can have written one).

**Decision: 🔴 Request changes** (agrees with the BLOCK verdict). Fixing and running Round 3 verification.

smarzban:
## Empanel — Round 3

Findings: 3 total — 0 critical · 0 high · 0 medium · 3 low · 0 info.
_Reviewed by:_ holistic · models: anthropic:claude-opus-5, openai-codex:gpt-5.6-terra

### Progress since Round 2
✅ Resolved (1): \[security\] Check the body cap before accepting completed curl output — not present in this round's findings
⏳ Still blocking (0)
🆕 New / regressed (0)

### Carry-forward: cleared by the verifier (1)
A `resolved` disposition with evidence cleared these prior gating claims:
- \[security\] Check the body cap before accepting completed curl output (`src/update/gateway.rs:459`)

### Advisory (non-blocking)
- ⚪ **[LOW]** \[correctness\] Over-cap re-check silently changes 404 semantics from conclusive-absent to unavailable — `src/update/gateway.rs:461` · 1/2 models · _by:_ holistic _(correctness)_
  The new \`over\_cap()\` re-check runs before the status bytes are classified, so a \`404\` whose response body exceeds \`DOCUMENT\_MAX\_BYTES\` now yields \`CaptureFailure::OverCap\` → \`Source::Unavailable\` instead of the previous \`Source::Available(None)\` → \`SpotlightInput::Missing\` → conclusive \`Withdrawn\`. A hostile/compromised source can therefore keep a stale spotlight alive (Preserve, no retrieval timestamp recorded) where it previously forced a withdrawal, which is a documented behaviour (\`ARCHITECTURE.md\`: "\`404\` withdraws a spotlight").
  _Fix:_ Either accept and note this deliberately (the over-cap body is not trustworthy evidence of anything), or move the re-check so it only gates the \`b"200"\` body-read arm in \`document\_with\_command\_with\_spawner\`, leaving a \`404\` conclusive regardless of the transient body's size.
- ⚪ **[LOW]** \[test-coverage\] No exact-boundary test for MAX\_TITLE\_LINE\_BYTES — `src/update/spotlight\_policy.rs:340` · 1/2 models · _by:_ holistic _(test-coverage)_
  Every other cap in this module is pinned on both sides (\`exact\_body\` at the cap is accepted, cap+1 withdraws; the title-char cap asserts an exact count). The new heading-line cap only has an over-cap case, so an off-by-one in the \`\>\` comparison (e.g. flipping to \`\>=\`) would silently reject a heading exactly at the limit without failing any test.
  _Fix:_ Add a case with a heading line of exactly \`MAX\_TITLE\_LINE\_BYTES\` bytes (\`"\# " + "T".repeat(MAX\_TITLE\_LINE\_BYTES - 2)\`) asserting it is still accepted.
- ⚪ **[LOW]** \[maintainability\] \`project\` doc-comment does not mention the new heading-line cap — `src/update/spotlight\_policy.rs:61` · 1/2 models · _by:_ holistic _(maintainability)_
  The \`project\` doc enumerates the exact acceptance contract ("valid UTF-8 whose first nonblank line is a \`\# \` heading with a nonempty visible title, and whose body fits \[\`MAX\_BODY\_BYTES\`\]") but the change adds a third rejection reason — a heading line over \`MAX\_TITLE\_LINE\_BYTES\` — that is now a conclusive withdrawal and is not documented at the one place callers read the contract.
  _Fix:_ Extend the \`project\` doc to state that a heading line longer than \`MAX\_TITLE\_LINE\_BYTES\` is a conclusive withdrawal, alongside the existing oversized-body rule.

smarzban:
## Orchestrator review — Round 3 (final) — ✅ Approve

**What the PR implements:** advisory remote notices — release details and a project spotlight fetched once daily from the fixed official repository (hardened `git ls-remote` + system `curl`), shown in a dismissable status row and the Help **What's New** section; display-only, fail-silent, cached in one atomic, safe-to-delete snapshot.

**This round:** verifier confirmed the Round-2 blocker **resolved** (the completed-transfer over-cap re-check); both fix-scoped seats reviewed the one-commit diff — terra clean, opus raised only low/info advisories; scan clean. Verdict: **pass**, nothing gating carried forward.

**Fixed across the gate (cumulative):**
- R1→R2: chunked/length-less downloads could bypass `--max-filesize` and write unbounded to the private body file — now bounded by a 50 ms size-watchdog that kills over-cap transfers mid-stream, with SECURITY.md corrected to describe the real bound.
- R1→R2: the What's New compose pre-pass ran on the input thread with only transport-level (1 MiB) bounds — remote content is now capped at the policy boundary (spotlight title 160 chars / heading line 4 KiB / body 64 KiB; release details 256 KiB), keeping Help-open cost proportionate to real content.
- R2→R3: a transfer completing between watchdog polls skipped the final cap check; the trim-before-truncate and heading-line-cap regressions in the title path.

**Deferred (advisory, logged):** char-vs-display-column title width; clearing over-cap cached details (unreachable for released builds); a dedicated test pinning the completion-time over-cap branch and the MAX_TITLE_LINE_BYTES boundary; one doc-comment enumeration and a latent rustdoc link. Plus the Round-1 advisories: Help-body recompose on mid-overlay refresh, temp-file orphans on quit-mid-refresh, session-start timestamps, one load-sensitive test.

**Suite:** verified green by the repo's CI required checks on every reviewed head (no green-bar config on this machine; the panel itself is read-only).

**Decision: ✅ Approve** (agrees with the `pass` verdict at head 54f5053).

PR #144, opened 2026-08-04
fix: Left/h steps up to the nearest visible ancestor when there's nothing to collapse

HewstonFox:
## Summary

`←`/`h` (the `collapse` action) no-oped whenever the selected row had nothing left to collapse: a **file**, or a directory that was **already collapsed**. This left the cursor stuck, with no way to keep climbing the tree via Left alone.

## Changes

- `src/controller/mod.rs`: `collapse()` now walks up to the nearest *visible* ancestor directory and collapses that instead of no-op'ing, mirroring how collapsing an expanded directory leaves it selected. Repeated presses climb the tree one level at a time. The walk climbs the real filesystem ancestry rather than stopping after one `parent()` hop, so it correctly skips past folded intermediate directories under `compact_dirs` (a folded chain like `src/main/java` is a single row keyed on its deepest directory — the immediate filesystem parent has no row of its own).
- `src/tree.rs`: adds `TreeModel::select(path)`, a small helper that repositions the cursor onto an already-visible row without expanding ancestors or relaxing filters (unlike `reveal`) — exactly what's needed right after collapsing an ancestor.
- `tests/controller.rs`: two new regression tests — one for the plain already-collapsed-directory case, one specifically for a `compact_dirs`-folded chain, to lock in that the walk climbs past every folded intermediate row.
- Docs: updated `docs/keys.md` and `docs/usage.md#the-tree` to describe the new walk-up behavior, plus a `CHANGELOG.md` entry under `[Unreleased]`.

## Test plan

- [x] `cargo test` is green
- [x] Docs updated in this PR (CHANGELOG + the relevant `docs/` page) for any user-facing change


smarzban:
thanks @HewstonFox for the PR and the feature.

PR #146, opened 2026-08-07
 feat: add configurable default view for changed files

umutciloglu:
## Summary

 Add an opt-in configuration setting for choosing the initial view of Git-changed files.

  The default remains diff-first for backward compatibility. Setting:

 ```toml
  changed_file_view = "content"
```
  opens changed Markdown in rendered mode and other changed files in syntax-content mode.

## Changes

  - Add changed_file_view = "diff" | "content" configuration.
  - Apply the preference before the initial render is dispatched.
  - Keep genuinely deleted files diff-first because they have no content to render.
  - Treat files that still exist after an index deletion, such as git rm --cached, as renderable content.
  - Preserve manual v cycling, full-diff behavior, status mode, filters, and baseline behavior.
  - Surface the effective value in the Settings overlay.
  - Update the configuration example, user documentation, changelog, and architecture map.
  - Add regression coverage for configured startup, deleted files, and on-disk replacements.

## Test plan

- [x] `cargo test` is green
- [x] Docs updated in this PR (CHANGELOG + the relevant `docs/` page) for any user-facing change

## Related

  N/A


smarzban:
Thanks @umutciloglu , this is a clean opt-in. I added a small config-to-startup test, and merged main so the PR is unblocked.

PR #151, opened 2026-08-16
feat: anchor launch decisions on the invocation context pane

diegopzz:
Makes the launchers honor the invocation context, so a *programmatic* invocation — `plugin.action.invoke` with an explicit `focused_pane_id`, e.g. a mirroring tool driving this host's viewer from another machine — opens the viewer beside the pane the caller named, instead of wherever the host's UI focus happens to sit.

## The problem

Both launch decisions anchor on the **focused pane**. That's right for a keybinding — the user is where the focus is. But when the action arrives with a context (herdr hands it to the action as `HERDR_PLUGIN_CONTEXT_JSON`, and callers like herdr-mirror's `remote-invoke` fill `focused_pane_id` deliberately, translated to this host's ids), the UI focus is unrelated: on a headless or remotely-driven host the viewer pops up in some arbitrary workspace, invisible to the caller. Concretely: invoking `open-file-viewer` through a mirror opened the viewer in whatever tab the host last had focused; with this change it opens beside the mirrored pane the caller invoked from — which then streams back to the mirroring side, making remote file browsing work end to end.

## What changes

- `context_pane()`: if the invocation context names a pane that exists in the `pane list`, that pane becomes the **anchor** for the whole decision (both variants). Absent, unparseable, or naming a dead pane → focused-pane behavior, unchanged.
- `OPEN` can now carry a target: `OPEN <pane_id>` (split variant → `--target-pane`), `OPEN <workspace_id>` (tab variant → `--workspace`). Only emitted under a context anchor, only after the same flag-safety validation as every other emitted id; bare `OPEN` stays the local default, so existing keybinding behavior is byte-identical.
- Under a context anchor, a Files pane already in the anchor's tab is `CLOSE`d (a repeat programmatic invocation means toggle — there is no meaningful "focused" state to flip to remotely).
- The binary reads `HERDR_PLUGIN_CONTEXT_JSON` itself (main.rs), keeping the scripts free of JSON parsing and the ids inside the tested/validated path, per the existing option-injection guard design.
- Scripts: one new `"OPEN "*` case each, passing the validated target through.

## Tests

Six new cases in `launch.rs` (context anchors OPEN with a target, repeat-toggle, dead-pane fallback, tab-variant workspace targeting, SWITCHTAB within the context workspace, garbage context changes nothing). `cargo test` (522), `cargo fmt --check`, and `cargo clippy --all-targets -- -D warnings` all green.

Verified live against herdr 0.8.0: a context naming a non-focused workspace's pane opens the viewer beside that pane; the same invocation again toggles it closed; and invoked on a mirrored host, the viewer pane streams back to the mirroring side and is fully drivable from there.

### DIFF
--- a/scripts/open-file-viewer-tab.sh
+++ b/scripts/open-file-viewer-tab.sh
@@ -25,11 +25,14 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
 viewer_bin="$script_dir/../target/release/herdr-file-viewer"
 
 open_tab() {
+  # extra args: `--workspace <id>` when the decision named an anchor — a
+  # context-carrying (programmatic) invocation opens the tab in THAT
+  # workspace, not the host's focused one. Validated by --launch-decision-tab.
   exec "$herdr_bin" plugin pane open \
     --plugin herdr-file-viewer \
     --entrypoint file-viewer \
     --placement tab \
-    --focus
+    --focus "$@"
 }
 
 decision="OPEN"
@@ -56,6 +59,9 @@ case "$decision" in
   "CLOSE "*)
     pid="${decision#CLOSE }"
     exec "$herdr_bin" pane close "$pid"
+    ;;
+  "OPEN "*)
+    open_tab --workspace "${decision#OPEN }"
     ;;
   *)
     open_tab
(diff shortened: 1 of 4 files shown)


smarzban:
Thanks for the contribution and clear explanation. Herdr 0.9 remote attach already supports the normal remote file-viewer workflow, so we do not need this extra context-targeting behavior at the moment. I’m closing this for now.

PR #152, opened 2026-08-18
feat: open_direction config key — split the viewer right (default) or down

diegopzz:
## What

A new config key deciding which way the launchers split the viewer pane off the invoking work pane:

```toml
open_direction = "right"   # the default — beside the work pane, today's layout
open_direction = "down"    # below it — suits tall/narrow layouts
```

The vocabulary is exactly what `herdr plugin pane open --direction` accepts, so the config can never request a placement herdr cannot perform.

## How

- `config.rs`: `OpenDirection` enum (`Right` default) resolved leniently (trimmed, case-insensitive) like `tree_position`; anything unrecognized keeps `Right` so a typo can never produce a `--direction` herdr rejects.
- The binary gains a third probe mode, `--open-direction`, printing the resolved label — the launchers cannot parse TOML, so they read the one value they need through the same defensive config path the TUI uses (missing binary / malformed config → `right`).
- `scripts/open-file-viewer.sh` and `scripts/open-file-viewer.ps1` consume the probe. The tab launcher is untouched — it has no split to aim.
- The read-only Settings overlay shows the effective value as its own row.
- Docs: key added to `docs/configuration.md` with the placement-keys paragraph extended.

## Tests

`cargo test` (1436 passed), `cargo fmt --check`, `cargo clippy --release --all-targets -- -D warnings` all clean. New coverage: resolver (config-wins / default / lenient), argv parsing (probe precedence), settings-overlay row.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -48,7 +48,7 @@ config key and above the built-in default — `editor` (`$EDITOR`) and `update_c
 (`$HERDR_FILE_VIEWER_NO_UPDATE_CHECK`) — giving those two a `config > env > default` chain. Every
 other key (`markdown`, `diff`, `syntax`, `open`, `reveal`, `hide_dotfiles`, `show_ignored`,
 `compact_dirs`, `changed_file_view`, `confirm_discard`, `scroll_lines`, `tree_width`,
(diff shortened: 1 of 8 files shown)


smarzban:
thanks @diegopzz I have merge the feature to main.

Issue #139, opened 2026-08-03
SKILL.md tells agents to pass --cwd, which always fails to spawn the viewer


AntonyKor:
Thanks for the detailed skill file — it's what made the viewer usable from an agent at all.

## What happened

`skills/herdr-file-viewer/SKILL.md` instructs agents to pass `--cwd` when opening the viewer:

> From the target repository or worktree, launch a fresh Files pane. Set `--cwd` to the agent's
> current working directory so the viewer resolves the correct worktree root.

```bash
herdr plugin pane open --plugin herdr-file-viewer --entrypoint file-viewer --placement split --direction right --cwd "$PWD" --focus --env "HERDR_FILE_VIEWER_OPEN=src/app.rs:42"
```

That command cannot succeed. `[[panes]].command` is the relative `./target/release/herdr-file-viewer`, and herdr resolves the program against `--cwd`, so the launch fails:

```
{"error":{"code":"plugin_pane_open_failed","message":"Unable to spawn /home/anton/projects/claude-inbox/./target/release/herdr-file-viewer because it does not exist"},"id":"cli:plugin"}
```

Dropping `--cwd` launches fine. This is the same spawn failure already root-caused in #88, but that issue was about worktree root resolution and was closed as fixed herdr-side in 0.7.3 — the `SKILL.md` instruction that triggers it was never updated, so an agent following the skill still hits it on 0.7.5.

## Steps to reproduce

1. From any git repo that is not the plugin checkout, run the `SKILL.md` command verbatim with `--cwd "$PWD"`.
2. Observe `plugin_pane_open_failed`.
3. Re-run the identical command without `--cwd` — the pane opens.

## Environment

- OS + version: CachyOS, Linux 7.1.5-1-cachyos
- herdr version: 0.7.5
- Plugin version: 1.14.0 (`fa1946a`)
- Renderers installed: glow 2.1.2, delta 0.19.2, bat 0.26.1
- Terminal: kitty 0.48.2, `TERM=xterm-256color`, inside herdr

## Suggested fix

Replace the `--cwd` instruction with the mechanism that actually determines the root, so agents stop reaching for a flag that breaks the spawn. Per `src/host.rs::parse_context`, the root comes from the injected `HERDR_PLUGIN_CONTEXT_JSON` (`focused_pane_cwd` → `workspace_cwd` → `cwd` → process cwd), resolved to the worktree top-level by `src/root.rs::resolve`. So the skill should say:

[...]

Issue #58, opened 2026-06-28
Feature Request: Windows Support


sanirudh17:
## Windows Support Request

Hello! First off, great project — a git-aware TUI file viewer with diffs, rendered markdown, and syntax highlighting is a really useful tool.

### Current Situation

According to the README and the repository metadata, `herdr-file-viewer` currently only supports **Linux** and **macOS**. Windows is not listed as a supported platform.

### Request

It would be greatly appreciated if Windows support could be added. Many developers work on Windows (natively or via tools like Windows Terminal), and having `herdr-file-viewer` available on Windows would make it accessible to a much wider audience.

### Possible Approaches

- **Native Windows support** via cross-compilation (the project is built in Rust, which supports Windows targets out of the box)
- **Windows Terminal / PowerShell compatibility** for the TUI rendering
- **WSL (Windows Subsystem for Linux)** as a short-term workaround could be documented if full native support is not immediately feasible

### Why This Matters

Since the project is written in Rust, adding Windows support may be relatively straightforward — Rust's cross-platform ecosystem (including `ratatui` and `crossterm`, which support Windows) makes this a realistic goal.

Would love to see this happen. Happy to help test on Windows if needed!

smarzban:
Hi @sanirudh17 — thanks for the thoughtful request, and for offering to test! 🎉

Good news: I'm  going to add native Windows support. The timing works out well, herdr itself runs natively on Windows (currently on its preview channel) and is actively improving it, and the viewer is Rust + ratatui/crossterm, which are already cross-platform. The gap is a small, known set of unix-only assumptions rather than anything architectural.

A few notes on how I'm approaching it:

- It'll ship labelled preview at first, mirroring herdr's own Windows status. On Windows you'll need herdr's preview channel (herdr channel set preview), and I'll prebuild a SHA-256-verified .exe so installing won't require a Rust toolchain — same as on Linux/macOS.
- In the meantime, WSL works today. herdr-inside-WSL is just herdr-on-Linux, and the existing Linux binary runs there unmodified — so if you'd like the viewer on a Windows machine right now, running herdr in WSL is a zero-wait option.
- I'd love to take you up on the test offer. Once there's a native Windows build to try, I'll ping this issue real-hardware validation is exactly what I need before promoting it past "preview."

Will keep this issue updated as the work lands. Thanks again for the nudge! 🙏

smarzban:
[...]

PR #158, opened 2026-08-24
feat: add open-file-viewer-overlay action

lexthink:
## Summary

Adds a third open action, `open-file-viewer-overlay`, that opens the viewer **over** the current pane instead of beside it (`open-file-viewer`) or in its own tab (`open-file-viewer-tab`).

It uses herdr's `overlay` placement, which is a split beside the active pane with the tab zoomed onto it: the viewer fills the tab, the layout underneath is untouched, and closing it puts the tab back. Useful for a quick look at a file without shrinking the pane you're working in. Today the only way to get this is a raw `[[keys.command]] type = "shell"` calling `herdr plugin pane open --placement overlay` by hand.

## Changes

- `scripts/open-file-viewer-overlay.sh` — launcher, same launch-or-focus-or-toggle shape as `open-file-viewer.sh`, scoped to the current tab. It reuses `--launch-decision` as-is (the decision only asks whether a `Files` pane exists in the focused pane's tab, so it doesn't care how that pane was placed — one viewer per tab across all three launchers, no new Rust). The one deliberate difference: FOCUS is `pane zoom <id> --on` alone. The split launcher follows it with `--off` to un-maximize back into the split, which here would flatten the overlay into a plain 50/50 split.
- `scripts/open-file-viewer-overlay.ps1` — Windows sibling. It can't ask herdr for `--placement overlay` (same absolute-path spawn reason as the other Windows launchers, #58), so it builds the state by hand: `pane split` + `pane run` + `pane rename Files` + `pane zoom --on`.
- `herdr-plugin.toml` — `open-file-viewer-overlay` (linux/macos) + `open-file-viewer-overlay-windows`, following the existing action pairs.
- Tests: `manifest.rs`, `launcher_content.rs`, `launcher_ps1.rs` extended to cover the new pair; new `tests/launcher_overlay.rs` pins that neither overlay launcher ever runs `zoom --off` and that the `.ps1` zooms the pane it splits; `docs_consistency.rs` adds the new script to the `--cwd` guard (#139) and a small anti-drift check that every `[[actions]]` id in the manifest has a `plugin_action` binding in `docs/summoning.md` / `docs/windows.md`.
- Docs: new section in `docs/summoning.md` (with a `prefix+alt+f` example), `docs/windows.md` action ids, README quick-start binding, `CHANGELOG.md` Unreleased entry.

No version bump — additive, and existing launchers are unchanged. The `overlay` placement has been available (and is the manifest default) since herdr 0.7.0, the manifest's `min_herdr_version`, per herdr's plugins doc at that tag, so the version floor doesn't move.

## Test plan

- [x] `cargo test` is green (full suite, including the new `launcher_overlay` tests)
- [x] `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings` pass (`cargo audit` not run locally — no dependency changes)
- [x] Docs updated in this PR (CHANGELOG + `docs/summoning.md`, `docs/windows.md`, README)
- [x] Verified live on macOS, herdr 0.8.2, with the release binary from this branch: OPEN opens the viewer zoomed and focused → repeat press CLOSEs it; with the viewer un-zoomed and unfocused in the same tab, the decision is FOCUS and `zoom --on` brings it back as the overlay → repeat press CLOSEs it.
- [ ] Windows: **not verified on real hardware.** The `.ps1` mirrors `open-file-viewer.ps1` plus the zoom step and is covered by the parse/content tests, but I only have a Mac.

## Related

Doesn't overlap with #151 / #152 (this PR only adds files and new manifest entries). If #151 lands first, the overlay launcher should pick up its `"OPEN <pane_id>"` case as a follow-up so all three launchers stay consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All notable changes to this project are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+- `open-file-viewer-overlay` action (`open-file-viewer-overlay-windows` on Windows): open the viewer as an overlay over the current pane — a split with the tab zoomed onto it, so the layout underneath is untouched and closing it puts the tab back. Toggles like the split action: a repeat press brings an unfocused viewer back up, or closes a focused one. Bind it to e.g. `prefix+alt+f`. → [summoning](docs/summoning.md#open-as-an-overlay-over-the-current-pane) · [windows](docs/windows.md)
+
 ## [1.16.0] - 2026-08-15
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -88,6 +88,12 @@ key = "prefix+shift+f"
 type = "plugin_action"
 command = "herdr-file-viewer.open-file-viewer-tab"
 description = "open file viewer in tab"
+
+[[keys.command]]
+key = "prefix+alt+f"
+type = "plugin_action"
+command = "herdr-file-viewer.open-file-viewer-overlay"
+description = "open file viewer as overlay"
 ```
 
 Run `herdr server reload-config`, then press your key. That's the whole setup: the split-pane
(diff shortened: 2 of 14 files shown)


smarzban:
Thanks @lexthink  for the implementation. We can already get this workflow with prefix+F (shift+f) to open Files in its own tab, or Z (shift+z) on the selected file for full-screen reading. I’m closing this for now.

PR #161, opened 2026-08-31
feat: session view — show the files the current Claude Code session touched

sxhmilyoyo:
## Summary

**Session view**: press `s` and the tree shows what the current Claude Code session **created (`+`), updated (`~`), or mentioned (`·`)** instead of the directory listing — the companion-pane experience of Claude Desktop's file panel, inside herdr. `s` again restores the full tree.

It reads the session's own transcript store (`~/.claude/projects/<slug>`), so there is nothing to install or configure. The view **follows the session live** (including Task-tool subagent transcripts) without ever moving your cursor, and files the session touched **outside the root** (a `~/.claude/CLAUDE.md` edit, a `/tmp` scratch file) list under an `── outside root ──` divider — preview-enabled but never browsable: navigation stays root-bound. `S` picks among the root's sessions (newest auto-followed by default; an explicit pick holds), `]`/`[` jump across members, and `session_view = true` opens straight into it. Works in non-git directories too.

Design notes for review:

- **Membership is deterministic only**: file-tool *results* (Write's `create`/`update` discriminator, Edit, Read), user `@`-mentions, and subagent transcripts (`<session-id>/subagents/*.jsonl`). Shell-command side effects are never inferred, and denied/failed calls never count — so a wrong entry can't erode trust in the panel.
- **Fail-soft against an undocumented format**: transcripts are a Claude Code internal, so parsing skips malformed/unknown lines, tail-reads incrementally in 4 MiB chunks per tick (a huge transcript streams in instead of blocking input), rebuilds cleanly when a file shrinks or vanishes, and degrades to an empty view with a notice — never a crash. SECURITY.md documents the transcript as a new untrusted input.
- **Root invariants preserved**: the in-root half reuses the changed-only synthesis (git markers, compaction, diff views all intact), and outside-root members waive exactly one guard — the content classifier's root-containment check, per-call and explicit — with every other protection (regular-file-only, bounded read, NUL check, escape neutralization) unchanged.
- **Zero impact outside the feature**: the glyph column and separator exist only while the view is active; every pre-existing insta snapshot is untouched.

## Changes

- New `session` module: transcript discovery (slug derived per UTF-16 code unit to match Claude Code's own mapping), an incremental multi-file `Follower` (main + subagent transcripts), strongest-category member map (`created ≻ updated ≻ mentioned`), root partitioning.
- `tree`: two-section session synthesis (shared emitter + separator + always-compacted, `~`-abbreviated outside groups), separator-aware cursor, expand/collapse on synthesized dirs and outside groups (rows start expanded; collapses survive live updates), `reveal` relaxation, member jump, missing-member `!` cue snapshotted once per set application (no per-frame stats).
- `controller/session`: the `s` toggle (mutually exclusive with `c`/`d`), throttled live-follow on `poll()` (500 ms appends / 3 s newest-rescan so two concurrent sessions can't thrash the view), session picker with explicit hold, worktree-switch reset, startup config.
- `presenter`: category glyph column, `[session]` title tag, divider row, session picker overlay; a member deleted since the session touched it keeps its row and gets a dedicated placeholder instead of `[binary file]`.
- Keys `s` / `S` (remappable as `session_view` / `session_picker`), config key `session_view`, Settings + Keybindings overlay rows.
- Docs in this PR: `docs/usage.md` (new section), `docs/keys.md`, `docs/configuration.md`, `config.example.toml`, `README.md` (including two screenshots of the session view and the session picker, captured from the real binary), `CHANGELOG.md` (Unreleased), `SECURITY.md`, `ARCHITECTURE.md`; both anti-drift key lists extended to cover the new key.
- 36 new tests across five layers: parser unit tests, transcript-file integration (append / partial line / shrink / vanish / subagents / chunked catch-up), tree synthesis, controller behavior over a fixture `$HOME`, presenter render assertions, and a pty e2e that toggles the view and previews an outside-root member end-to-end.

## Test plan

- [x] `cargo test` is green — 1,473 passed, 0 failed (includes the new pty e2e)
- [x] `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings` pass; no new dependencies, so `cargo audit` is unchanged from `main`
- [x] Docs updated in this PR (CHANGELOG + the relevant `docs/` pages) for the user-facing change

## Related

The two decisions behind the feature (transcript parsing over a hook ledger; the outside-root display-not-browse carve-out) are written up as ADR-0011/0012, continuing the repo's local ADR numbering — the tracked citations in `SECURITY.md`/`ARCHITECTURE.md`/source comments refer to those.


smarzban:
Thanks @sxhmilyoyo for the substantial work on this. I’m keeping the viewer agent-neutral and scoped to the selected project tree, so Claude-specific transcript parsing and outside-root session files are not a fit here. Closing this for now.

PR #167, opened 2026-09-15
feat: open_direction config key — split the viewer below instead of beside

pmaxvsbobo:
## What

Adds an `open_direction` config key so the summon action can split the viewer **below** the invoking pane instead of beside it:

```toml
open_direction = "down"   # or "right" (the default); "bottom" is accepted too
```

`prefix+f` currently always splits to the right — `scripts/open-file-viewer.sh` hardcoded `--direction right`. On a narrow terminal that leaves the viewer under `preview_layout::NARROW_SPLIT` (80 cols), so it collapses to a single column and you only see the tree. A bottom split gives the viewer full width.

## How

- New `config::OpenDirection` (`Right` / `Down`) + `open_direction: Option<String>`, resolved leniently (trim, case-insensitive, `bottom` as a synonym for `down`) exactly like `tree_position`.
- New `--open-direction` CLI action prints the effective value. The launchers probe the **binary**, so the lenient-value rules live in one tested place rather than being re-implemented in bash and PowerShell, and the shell re-validates the token before it reaches an argv (same option-injection posture as the pane ids).
- Both split launchers resolve `HERDR_PLUGIN_CONFIG_DIR` themselves via `plugin config-dir` first. herdr injects that variable into a pane it spawns from the manifest, **not** into an action's process — without this the probe reads the standalone `$XDG`/`$HOME` path and silently misses the user's config. This is the same trap already documented in `open-file-viewer.ps1`'s `Get-ConfigDir`.
- The probe runs on the **OPEN** path only; FOCUS/CLOSE are untouched.

## Scope and compatibility

- **Split action only.** The tab launcher is deliberately untouched (a tab has no direction), and the manifest keeps `placement = "split"` per AC-17 — `tests/manifest.rs` still passes unchanged.
- **Default is unchanged.** Any failure — binary missing, config unreadable, unrecognized value — degrades to `right`, so an existing install behaves identically until the key is set.
- Read at summon time, so it applies to the next viewer opened, not to one already on screen. No reload needed.

## Verified against herdr

`right` and `down` are the only values herdr 0.9.0 accepts, confirmed live:

```
$ herdr plugin pane open --help
      --direction <DIRECTION>   [possible values: right, down]
$ herdr pane split --help
      --direction <DIRECTION>   [possible values: right, down]
```

`OpenDirection::label` is pinned to that vocabulary by `open_direction_labels_are_herdr_direction_tokens`, with the verified argv in the test comment.

## Drift guard

`split_launchers_take_the_direction_from_the_open_direction_probe` (`tests/docs_consistency.rs`) fails the build if either launcher re-hardcodes `--direction right`. Without it the `Config` field, the docs, the `?` Settings row and the resolver tests could all survive while the setting itself became a silent no-op — the same failure mode the `--cwd` guard (#139) was written for.

## Docs (same PR)

`CHANGELOG.md`, `docs/configuration.md`, `docs/summoning.md` (new "Split beside or below"), `config.example.toml`, the `?` Settings row, and `ARCHITECTURE.md`.

## Version

Bumped to **1.17.0** in `Cargo.toml`, `Cargo.lock` and `herdr-plugin.toml`. This changes how a launcher script invokes the binary, which AGENTS.md requires be bumped in the same commit rather than batched into a later release. Happy to drop the bump if you'd rather take it through a `release/` PR.

---

⚠️ **Not built or tested locally** — the machine I authored this on has no Rust toolchain and no working container runtime, so I could not run `cargo fmt --check`, `clippy`, or the suite even once. Everything was written against the surrounding patterns and reviewed by hand, but please treat CI as the first real signal. Happy to fix up whatever it turns red on.


smarzban:
thanks @pmaxvsbobo for the PR

